Issue #7, opened 2018-08-14
koa-bodyparser


kakoc:
I need to parse multipart/form-data but current bodyparser(koa-bodyparser) can't do that.
https://github.com/koajs/bodyparser/issues/85
I found koa-body. This bodyparser in addition can do it. Maybe we should replace main bodyparser or add additional middleware for parsing form-data?
If you agree, i created PR #8

PR #8, opened 2018-08-14
Bodyparser middleware

kakoc:
PR for fixing issue #7 if you are agree.

### DIFF
--- a/app/index.ts
+++ b/app/index.ts
@@ -1,7 +1,7 @@
 import * as cors from '@koa/cors';
 import { createLogger } from 'bunyan';
 import * as Koa from 'koa';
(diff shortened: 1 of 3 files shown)


R1ZZU:
Seems like bodyparser middleware from koajs org is also capable of handling form submit

https://github.com/koajs/bodyparser/blob/db193f5d46860393521ad38f90a554968b2ba98a/index.js#L53

kakoc:
You point to 'application/x-www-form-urlencoded', but i mean 'multipart/form-data'(i use it for sending files)

R1ZZU:
Consider other options of sending files (blob/base64)

PR #9, opened 2018-08-18
Batch update

kakoc:
Need implement Assignments model and working with it for successful integration of butchUpdate.

I am ready get this task.


app/routes/user/match.ts
```diff
@@ -0,0 +1,23 @@
...
+import { userService } from '../../services';
+import { Roles, IUserModel } from './../../models/user';
+import { setResponse, toArray } from '../utils';
+
+export const matchUsers = async (ctx: Router.IRouterContext) => {
+    const {
+        data: { role, forCheck },
+    } = ctx.request.body;
+
+    const users = await userService.getBunchUsers(toArray(forCheck), Roles[role] as Roles);
```
Shastel: I'm sure, that you can extract directly id from DB

kakoc: i think that we don't have to work with db directly in our routes. in the future we can have more complex logic for working with data and i think it's a good idea use some services which will encapsulate a complex logic or just services which would be used as collection managers. 
if i right understand you.

PR #17, opened 2018-08-31
Student Tasks Stage1(Assignments)

KurnosovNikita:


app/routes/course/assignments.ts
```diff
@@ -0,0 +1,120 @@
...
+        ctx.status = OK;
+    } catch (err) {
+        ctx.logger.error(err);
+        ctx.status = INTERNAL_SERVER_ERROR;
+    }
+};
+
+export const getNormalizeAssignmentsData = (assignments: AssignmentResponse[]): NormalizeAssignmentsData => {
+    const sortedAssignments = assignments
+        .reduce<INormalizeAssignments[]>((res, assignment) => {
```
yuliaHope: It would be better to use `map` here and move some calculations to util function.

app/routes/course/assignments.ts
```diff
@@ -0,0 +1,120 @@
...
+            const a = assignmentA.assignment.startDateTime!;
+            const b = assignmentB.assignment.startDateTime!;
+            return b - a;
+        })
+        .sort((assignmentA, assignmentB) => {
+            const a = Number(assignmentA.isEndAssignment!);
+            const b = Number(assignmentB.isEndAssignment!);
+            return a - b;
+        });
+    const data = sortedAssignments.reduce<NormalizeAssignmentsData>(
```
yuliaHope: Why reduce here??? You can just return `{ assignments: sortedAssignments }`.

app/routes/course/assignments.ts
```diff
@@ -0,0 +1,120 @@
...
+                res.push({ assignment, isEndAssignment: false, isActiveAssignment: true });
+            }
+            return res;
+        }, [])
+        .sort((assignmentA, assignmentB) => {
+            const a = assignmentA.assignment.startDateTime!;
+            const b = assignmentB.assignment.startDateTime!;
+            return b - a;
+        })
+        .sort((assignmentA, assignmentB) => {
```
yuliaHope: I think it would be better to use lodash sortBy here to sort by two params. https://lodash.com/docs/4.17.10#sortBy

app/routes/course/assignments.ts
```diff
@@ -0,0 +1,120 @@
...
+            if (assignments[index]) {
+                arrayOfTasksId.push(assignments[index].taskId);
+            }
+        }
+        const tasks: IEventModel[] = await getTasksByTaskId(arrayOfTasksId);
+        const combinedAssignmentsAndTasks: AssignmentResponse[] = [];
+        for (const index in assignments) {
+            if (assignments[index] && tasks[index]) {
+                combinedAssignmentsAndTasks[index] = { ...Object.create(null), ...tasks[index], ...assignments[index] };
+            }
```
yuliaHope: Please, rewrite this part. I think it would be better to use  `populate` or some aggregation here.
Docs: https://mongoosejs.com/docs/populate.html

app/routes/event/Events.ts
```diff
@@ -0,0 +1,158 @@
...
+    await AssignmentModel.bulkWrite(assignments);
+};
+
+export const createDeleteEventsRoute = async (ctx: Router.IRouterContext) => {
+    const { id, eventType } = ctx.params;
+    try {
+        const query =
+            eventType === 'session' ? await SessionModel.findByIdAndRemove(id) : await TaskModel.findByIdAndRemove(id);
+        if (query !== null) {
+            await AssignmentModel.remove({ taskId: id });
```
yuliaHope: Do you try to remove assignment in case deleting session?

app/routes/event/Events.ts
```diff
@@ -0,0 +1,158 @@
...
+        ctx.status = OK;
+    } catch (err) {
+        ctx.logger.error(err);
+        ctx.status = INTERNAL_SERVER_ERROR;
+    }
+};
+
+const getStudentsByCourseId = async (courseId: string) => {
+    const result: ICourseStudent[] = await CourseStudentModel.aggregate([
+        {
```
yuliaHope: You needn't do aggregation here. To get userId and mentors you can use `CourseStudentModel.find({courseId})`

app/routes/assignment/assignment.spec.ts
```diff
@@ -0,0 +1,63 @@
+import { IRouterContext } from 'koa-router';
+import { AssignmentStatus } from '../../models';
+import { INTERNAL_SERVER_ERROR, OK } from 'http-status-codes';
+import { setResponse } from '../utils';
+
+const createPatchAssignmentRoute = async (ctx: any, next: any) => {
+    ctx.body.status = AssignmentStatus.Checked;
```
yuliaHope: You shouldn't create your own functions and test them. You should test `createPatchAssignmentRoute ` from `app/routes/assignment/assignment.ts`. I mean that in your test cases you should call `createPatchAssignmentRoute `, but before you should mock AssignmentModel.
For Example:
```js
import { createPatchAssignmentRoute } from './assignment';
import { AssignmentModel} from '../../models';

jest.mock('../../models', () => ({
    AssignmentModel: { findByIdAndUpdate : jest.fn().mockReturnValue(someObjectHere) },
}));

describe('app/routes/assignment/assignment.ts', () => {
    it('returns 200 if returns data', async () => {
        const ctx = createTestContext();
        await createPatchAssignmentRoute(ctx, findByIdAndUpdate);
        expect(AssignmentModel.findByIdAndUpdate).toHaveBeenCalledTimes(1);
        expect(ctx).toMatchSnapshot();
    });
```

yuliaHope: The same for all tests

app/routes/course/assignments.ts
KurnosovNikita: Fixed.

app/routes/course/assignments.ts
KurnosovNikita: Fixed.

app/routes/course/assignments.ts
KurnosovNikita: Fixed with orderBy.

app/routes/event/Events.ts
KurnosovNikita: Yes, fixed.

app/routes/event/Events.ts
KurnosovNikita: Fixed.

app/routes/course/assignments.ts
KurnosovNikita: It is impossible to implement this part with *aggregation*,  because we have to compare taskId(String) and _id(ObjectId) and there is no way to convert them to the same type(an example is given below). Tried with : _id.str , { "$toString": "$_id" }, $convert. We can try to change the type of "taskId" in the Assignment collection to ObjectId, but I don't think it would be a good idea.
```javascript
const assignments = await AssignmentModel.aggregate([
{
$match: {courseId,studentId}
},
{
$lookup:{
from: 'tasks',
localField: 'taskId', // String
foreignField: '_id', // ObjectId
as 'task'
}
},
{
$unwind: '$task'
},
])
```
Perhaps this can be done with *populate*, but so far I have not found a way to implement it.

PR #67, opened 2019-07-01
/registry route: GET, PUT, POST

anastasiaprivalova:


src/models/registry.ts
```diff
@@ -0,0 +1,29 @@
...
+export class Registry {
+  @PrimaryGeneratedColumn() id: number;
+
+  @ManyToOne(_ => User, (user: User) => user.registries, { nullable: true })
+  user: User | number;
+
+  @ManyToOne(_ => Course, (course: Course) => course.registries, { nullable: true })
+  course: Course | number;
+
+  @Column({ name: 'type', enum: ['mentor', 'mentee'] })
```
apalchys: please don't use `enum` type. We discovered some issues related to integration with other systems (Kafka) and moving way from them.

PR #91, opened 2019-08-19
feat: github task checker

daniil-loban:
2.1.  + Download repository (master branch) as zip and unpack
2.2.  + Install deps
2.3.  + Run "npm test"
2.4.  - Save result to DB
 
**the return result from the function looks like this**:
```
{
  "statusCode": 200,
  "body": {
    "passed": 100,
    "failed": null,
    "time": "40ms",
    "exitCode": 0
  }
}
```

### DIFF
--- a/serverless/github-task-checker/handler.ts
+++ b/serverless/github-task-checker/handler.ts
@@ -1,7 +1,163 @@
 import { APIGatewayProxyHandler } from 'aws-lambda';
-import 'source-map-support/register';
+//import 'source-map-support/register';
+
+import * as fs from 'fs';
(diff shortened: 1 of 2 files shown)


serverless/github-task-checker/handler.ts
```diff
@@ -1,16 +1,96 @@
...
+        }
+        if (failed) {
+          result.failed = +failed[1];
+        }
+      });
+    });
+  });
+};
+
+const worker = (event): Promise<string> => {
```
apalchys: please add interface/type to event which will declare expected input format.

PR #114, opened 2019-09-10
feat: add statistics, cities & from EPAM flag for registration dashboard

anastasiaprivalova:


client/src/pages/admin/registrations.tsx
```diff
@@ -70,6 +77,23 @@ class RegistrationsPage extends React.Component<Props, State> {
...
+        case 'rejected':
+          statistics.rejected += 1;
+          break;
+        case 'pending':
+          statistics.pending += 1;
+          break;
+      }
+    }
+
+    console.log(statistics);
```
apalchys: please remove :)

client/src/pages/admin/registrations.tsx
```diff
@@ -112,11 +145,11 @@ class RegistrationsPage extends React.Component<Props, State> {
   };
 
   approve = (e: any) => {
-    this.handleSubmit(e, 'approve');
+    this.handleSubmit(e, 'approved');
```
apalchys: i believe you also need to updated the string on the server side. 
https://github.com/rolling-scopes/rsschool-app/blob/master/server/src/routes/registry.ts#L132

client/src/pages/admin/registrations.tsx
anastasiaprivalova: Thanks for the note, fixed

PR #136, opened 2019-09-25
open github from score

lesnitsky:


Shastel:
How to open profile in a new tab then?

lesnitsky:
maybe worth opening profiles in new tab by default then?

Shastel:
I'm doing approx 50/50, but we should gather more feedback from someone who probably opens a lot of profiles? 😃
cc @dzmitry-varabei

apalchys:
i vote for opening external links (github.com) in a new tab by default.

lesnitsky:
The problem is that this link leads to two different locations:

1) https://app.rs.school/profile?githubId=<githubId> <https://app.rs.school/profile?githubId=%3CgithubId%3E>
2) https://github.com/<githubId> <https://github.com/%3CgithubId%3E>

Github seems to be secondary locations, so the idea is to open it with click modifier (meta or ctrl)
These modifiers default to "open link in new tab" though, so proposal is

1 opens in new tab by default 
2 opens github profile in new tab with ctrl/meta

Shastel:
@apalchys external for sure, but we talking about `/profile?githubId=` page

apalchys:
sorry for confusion. did check the PR carefully.

dzmitry-varabei:
@lesnitsky @Shastel @apalchys Hi folks!
My two cents. Since we have two columns in the score - "Github" and "Name", we can use "Github" to open the Github profile and "Name" to open the RSAPP profile. For me, this behavior is more clear then the usage of click modifier

lesnitsky:
@dzmitry-varabei implemented as you suggested
@apalchys could you please review?

PR #142, opened 2019-10-01
feat: add possibility for scoring multiple students

mikhama:


client/package-lock.json
```diff
@@ -1,8804 +0,0 @@
-{
```
apalchys: please keep this file. I added it recently to fix an issue with building Docker image. 
It's not easy to integrate existing yarn setup with how we build our images and in the same time, we don't want to do fresh install via npm during Docker build step. (i faced with multiple issues related to unexpected package updates which led to downtime:))

client/src/next.config.js
```diff
@@ -14,6 +14,7 @@ const nextConfig = {
   webpack: config => {
     config.resolve.alias['components'] = path.join(__dirname, 'components');
     config.resolve.alias['services'] = path.join(__dirname, 'services');
+    config.resolve.alias['utils'] = path.join(__dirname, 'utils');
```
apalchys: 👍

PR #312, opened 2020-01-27
Add Elista to list of cities

dzhudzhi:
Give Elista id - 14454 and parentId - 10191


apalchys:
@dzhudzhi sorry for delay with adding Elista. 

There was a plan to fix this issue (and similar) with migration to a city DB from `geonames.org` and moving hardcoded cities from json to applications DB. Unfortunately it takes more than expected.
For now, I pushed a "hotfix" to master with Elista and id = `563514` (from geonames.org). 
Let me know if you have questions.  Thanks for your PR!

PR #339, opened 2020-02-25
fix: profile access for managers/coordinators

apalchys:


server/src/routes/profile/permissions.ts
```diff
@@ -161,32 +169,55 @@ export const getPermissions = ({ isAdmin, isProfileOwner, role, permissions }: P
...
+        'isProfileVisible',
+      ].includes(permission) && ['mentor', 'coursementor', 'coursemanager'].includes(role)
+    );
+  };
+
+  const accessToProfile = (permission: string, role: string = '') =>
+    ['isProfileVisible'].includes(permission) && ['student'].includes(role);
+
+  return mapValues(defaultPermissions, (_, permission) => {
+    if (isAdmin || role === 'coursemanager') {
```
apalchys: i changed this function and made it more verbose but simpler for future modifications. initially it was hard to understand what condition does.

PR #395, opened 2020-04-04
Feat/notifications

NiksRakevich:


server/src/services/consentService.ts
```diff
@@ -0,0 +1,27 @@
+import { Consent } from '../models/consent';
+import { getRepository, In } from 'typeorm';
+
+export async function captureConsent(consent: Consent) {
```
apalchys, line 4: recently I decided to start work with DB normally and clean the mess that we have  right now.  Idea is to   adopt using repository pattern for DB access. Example here: `rsschool-app/server/src/repositories/interview.ts`

Would be great if you convert this service to `repository`

server/src/repositories/consent.ts
```diff
@@ -0,0 +1,32 @@
...
+    const existingConsent = await this.getConsent(consent.channelValue);
+    const consentId = existingConsent?.id;
+    if (!consentId) {
+      await repository.insert(consent);
+    } else {
+      await repository.update(consentId, consent);
+    }
+  }
+
+  private getConsent(channelValue: string): Promise<Consent | undefined> {
```
apalchys, line 29: would be great to use `find` verb only for get operations

PR #447, opened 2020-06-11
Feature: add student dashboard

Pulya10c:
Create a page with a set of widgets which display all useful information for a student at glance.

Implement the widgets:
- Score
- Mentor
- Tasks statistics
- Next event

### DIFF
--- a/client/src/pages/index.tsx
+++ b/client/src/pages/index.tsx
@@ -14,6 +14,7 @@ import {
   StarOutlined,
   QuestionCircleTwoTone,
   CheckSquareOutlined,
+  DashboardTwoTone,
 } from '@ant-design/icons';
 import { Button, Card, Col, Layout, List, Result, Row, Select, Statistic, Tag, Typography } from 'antd';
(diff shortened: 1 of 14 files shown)


client/src/pages/index.tsx
```diff
@@ -67,6 +68,15 @@ const combineOr = (...checks: any[]) => (course: Course, role: Role, session: Se
...
 const routes = [
+  {
+    name: () => (
+      <>
+        <DashboardTwoTone /> Dashboard
+      </>
+    ),
+    getLink: (course: Course) => `/course/student/dashboard?course=${course.alias}`,
+    access: combineAnd(isCourseNotCompleted, isStudent),
+  },
```
apalchys, line 79: please remove it. Let's make it accessible only by direct link for now. When we test it and add all required functionality, then we will enable it on the home page

Pulya10c: @apalchys  
Thanks, I fixed issue

PR #463, opened 2020-07-06
feat: improve clearance of form fields  on submit

AlreadyBored:
Submit now clears all fields except the field with the task.


apalchys:
thanks!

PR #470, opened 2020-07-11
Fix self expel presence on public profile

AlreadyBored:
SelfExpel no longer appears on the profiles, the owner of which is not the user who is viewing themю

### DIFF
--- a/client/src/components/Profile/StudentStatsCard.tsx
+++ b/client/src/components/Profile/StudentStatsCard.tsx
@@ -15,6 +15,7 @@ const { Text } = Typography;
 type Props = {
   data: StudentStats[];
   isEditingModeEnabled: boolean;
+  isProfileOwner: boolean | null;
   permissionsSettings?: ConfigurableProfilePermissions;
   onPermissionsSettingsChange: (event: CheckboxChangeEvent, settings: ChangedPermissionsSettings) => void;
   username: string;
@@ -76,7 +77,7 @@ class StudentStatsCard extends React.Component<Props, State> {
   }
 
   render() {
-    const { isEditingModeEnabled, permissionsSettings, onPermissionsSettingsChange } = this.props;
+    const { isEditingModeEnabled, permissionsSettings, onPermissionsSettingsChange, isProfileOwner } = this.props;
     const stats = this.props.data;
     const gitHubId: string = this.props.username;
     const { isStudentStatsModalVisible, courseIndex, coursesProgress, scoredTasks } = this.state;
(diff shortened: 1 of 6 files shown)


client/src/components/Profile/StudentStatsCard.tsx
```diff
@@ -146,7 +147,7 @@ class StudentStatsCard extends React.Component<Props, State> {
                     <p style={{ fontSize: 12, marginBottom: 5 }}>
                       Score: <Text mark>{totalScore}</Text>
                     </p>
-                    {!(isExpelled || isCourseCompleted) ? (
+                    {!(isExpelled || isCourseCompleted || !isProfileOwner) ? (
```
DimaRomaniuk, line 150:
```suggestion
                    {!(isExpelled || isCourseCompleted) || isProfileOwner ? (
```

client/src/components/Profile/StudentStatsCard.tsx
```diff
@@ -15,6 +15,7 @@ const { Text } = Typography;
 type Props = {
   data: StudentStats[];
   isEditingModeEnabled: boolean;
+  isProfileOwner: boolean | null;
```
DimaRomaniuk, line 18: Why do you need null ?

AlreadyBored: To display in a state that profile ownership wasn't cheked because the profile wasn't found. I think this is different from `false`. But perhaps this is not necessary

client/src/components/Profile/StudentStatsCard.tsx
AlreadyBored: So it looks better

client/src/pages/profile/index.tsx
```diff
@@ -208,13 +210,20 @@ export class ProfilePage extends React.Component<Props, State> {
     try {
       const githubId = router.query ? (router.query.githubId as string) : undefined;
       const profile = await this.userService.getProfileInfo(githubId);
+      let isProfileOwner;
```
apalchys, line 213: for now, as we don't have any additional requirements, I would do it simple true/false, without undefined.

```
         const userId = this.props.session.githubId;
         const profileId = profile.generalInfo!.githubId;
         const  isProfileOwner = checkIsProfileOwner(userId, profileId);
```

client/src/components/Profile/StudentStatsCard.tsx
```diff
@@ -146,7 +147,7 @@ class StudentStatsCard extends React.Component<Props, State> {
                     <p style={{ fontSize: 12, marginBottom: 5 }}>
                       Score: <Text mark>{totalScore}</Text>
                     </p>
-                    {!(isExpelled || isCourseCompleted) ? (
+                    {!(isExpelled || isCourseCompleted) && isProfileOwner ? (
```
apalchys, line 150: suggest to change to the following line. It will require less mental work to evaluate condition (at least for me)

```suggestion
                    {(!isExpelled && !isCourseCompleted) || isProfileOwner ? (
```

AlreadyBored: Why not

PR #579, opened 2020-11-03
fix: various bugfixes

apalchys:


server/src/services/score.service.ts
```diff
@@ -0,0 +1,113 @@
...
+    .leftJoin('mentor.user', 'mu')
+    .addSelect(getPrimaryUserFields('mu'))
+    .where('student."courseId" = :courseId', { courseId });
+
+  if (filter.activeOnly) {
+    query = query.andWhere('student."isFailed" = false').andWhere('student."isExpelled" = false');
+  }
+
+  if (filter.name) {
+    query = query.andWhere('("user"."firstName" ILIKE :searchText OR "user"."lastName" ILIKE :searchText)', {
```
apalchys, line 59: fix is to add brackets @DmitryRomaniuk

DimaRomaniuk: thx

client/src/pages/profile/index.tsx
```diff
@@ -311,8 +311,10 @@ export class ProfilePage extends React.Component<Props, State> {
   };
 
   async componentDidMount() {
-    await this.fetchData();
-    await this.authorizeDiscord();
+    setTimeout(async () => {
```
DimaRomaniuk, line 314: Why do you use here setTimeout ?

apalchys: Thanks. I'll add a comment. It's a hack to fix some next.js caveat that router.query params are not available immediately on the client.

PR #716, opened 2021-04-27
fix the size of the table with tasks on the student's dashboard

### LINKED ISSUE #669: [https://app.rs.school/registry/mentor] Sign up for a specific course
If a mentor signs up for a specific course via a direct link(eg. https://app.rs.school/registry/mentor?course=js-fe-lt), the specific course should be checked by default in the "preferred course" section.

image

**Expected behavior**:
image

Pulya10c:
This fixes #669

image


akool:
This commit breaks the table display on desktops. It sets the width too narrow for columns "**Score / Max**" and "***Weight**".

PR #741, opened 2021-05-12
Новый алгоритм распределения на кроссчеке

Fodin:
Однопроходный алгоритм распределения без коллизий.
Поэкспериментировать можно здесь: https://codepen.io/fodin/pen/rNyxWxR
Работает по принципу шифра Цезаря.


apalchys:
@Fodin  Thanks! could you please run `npm run prettier` in your PR?

Fodin:
@apalchys Fixed. Sorry for commit name. I did it online and didn't find how to rename it.

Fodin:
@apalchys There is a potential infinite loop that can happens with the very rare cases of synthetic or empty input data, as example ```createCrossCheckPairs([1, 2, 3, 4, 5, 6], 3)``` will be looped but ```createCrossCheckPairs([1, 2, 3, 4, 5, 6, 7], 3)``` will not. It happens because impossible to create needed pairs without collisions when two students check each other.
I can write a check for it but I need to know how I can handle this case. Throw an error? Return empty array of pairs?

apalchys:
@Fodin it's better to throw an error. Thanks!

PR #769, opened 2021-06-01
fix: scrolling

apalchys:


ThorsAngerVaNeT:
@apalchys I have made an investigation of the disappearing scrollbar problem.

GIF 04 04 2022 15-45-40

That problem caused many students questions like "Why don't I have Git Test/Codewars?".
I find out that this commit has disabled one of the solutions that was in main.css. Another solution is to set property `virtual` to false for Select antd component.

PR #818, opened 2021-07-24
Add mentor ability to unassign student

AlreadyBored:


client/src/services/models.ts
```diff
@@ -31,3 +31,5 @@ export interface CoursePageProps {
   session: Session;
   course: Course;
 }
+
+export type Action = 'expel' | 'unassign';
```
valerydluski, line 35: can transfer to expelStudent.
or change the name of not just Action, but for what Action

AlreadyBored: done

valerydluski (approved):

PR #825, opened 2021-08-03
Feature: find bad student's reviews for crosscheck

KostyaDedulya:
image
1) Select who gave a comment for crosscheck and the comment was the first and his length was less than 70 symbols
2) Select who gave the same marks to all students and student don't have average score in range (0.9 * checker's score) and (1.1 * checkers score)


client/src/components/BadReview/BadReviewControllers.tsx
```diff
@@ -0,0 +1,112 @@
...
+            }
+          } catch (error) {
+            message.error('Something went wrong');
+          }
+          cache.current = { ...cache.current, [taskId]: { ...cache.current[taskId], [checkType]: dataFromService } };
+          setIsLoading(false);
+          setData(dataFromService);
+        }
+      }
+    }
```
mikhama, line 74: Wow... can we simplify it?

server/src/services/check.service.ts
```diff
@@ -0,0 +1,96 @@
+import { getRepository } from 'typeorm';
+import { CourseTask, Student, Task, TaskSolutionResult, User } from '../models';
+
+const LOW_ERROR_RATE = 0.9;
+const HIGH_ERROR_RATE = 1.1;
+const TASK_REVIEW_COUNT = 4;
+const MIN_LENGTH_MESSAGE = 70;
```
mikhama, line 7: Probably for future implementation.
But I think it will be fine to setup such variables from the UI.
What do you think?

server/src/services/check.service.ts
```diff
@@ -0,0 +1,96 @@
...
+    .addSelect('ts.score', 'checkerScore')
+    .innerJoin(
+      qb => {
+        return qb
+          .subQuery()
+          .select('tsr."studentId"')
+          .addSelect('AVG(tsr.score)', 'avg')
+          .from(TaskSolutionResult, 'tsr')
+          .groupBy('tsr."studentId"');
+      },
```
mikhama, line 26: just `qb => qb.subQuery...`

server/src/services/check.service.ts
```diff
@@ -0,0 +1,96 @@
...
+import { CourseTask, Student, Task, TaskSolutionResult, User } from '../models';
+
+const LOW_ERROR_RATE = 0.9;
+const HIGH_ERROR_RATE = 1.1;
+const TASK_REVIEW_COUNT = 4;
+const MIN_LENGTH_MESSAGE = 70;
+
+/* Get checkers who passed max score for everyone and maybe didn't review task*/
+export async function getCheckersWithMaxScore(taskId: number) {
+  const data = await getRepository(TaskSolutionResult)
```
mikhama, line 11: Really huge query...
Did you test it on some "big" data? It is work quite fast?

client/src/components/BadReview/BadReviewControllers.tsx
mikhama: Like move some part to the checkService?
Responsibility of "getData" should be only to get some data...

KostyaDedulya: +

server/src/services/check.service.ts
KostyaDedulya: It'll be in future )

server/src/services/check.service.ts
KostyaDedulya: +

server/src/services/check.service.ts
KostyaDedulya: I didn't test( because i didn't have big data
I asked for snapshot from prod app, but i didn't get

mikhama: I think you can use some tool for generating random data like this (https://www.mockaroo.com/).
It quite configurable.
Just fill tables that you need with some random data (I think max of 10000 students will be enough).
And then run it locally to check (at least use console.time).

If you need some help with generation random data - we can make some short call, just let me know

client/src/components/BadReview/BadReviewControllers.tsx
```diff
@@ -0,0 +1,110 @@
...
+        setIsLoading(true);
+        let dataFromService: IBadReview[] = [];
+        try {
+          switch (checkType) {
+            case 'Bad comment':
+              dataFromService = await checkService.getBadComments(taskId, checkType);
+              break;
+            case 'Did not check':
+              dataFromService = await checkService.getMaxScoreCheckers(taskId, checkType);
+              break;
```
mikhama, line 62: Better make it as a part of `checkService`

mikhama: Method that should call getBadComments or getMaxScoreCheckers depends on type...

client/src/components/BadReview/BadReviewControllers.tsx
```diff
@@ -0,0 +1,110 @@
...
+            case 'No type':
```
mikhama, line 63: and instead of that

```
default:
  throw new Error('Unsupported type');
```

client/src/services/check.ts
```diff
@@ -0,0 +1,26 @@
...
+  private axios: AxiosInstance;
+  private cache: Record<number, Record<checkType, IBadReview[]>>;
+
+  constructor() {
+    this.axios = globalAxios.create({ baseURL: `/api/` });
+    this.cache = {};
+  }
+
+  async getBadComments(taskId: number, type: checkType): Promise<IBadReview[]> {
+    if (this.cache?.[taskId]?.[type]) return this.cache[taskId][type];
```
mikhama, line 14: You can create separate function-wrapper that will create / return cache.
Just look at these functions: "getBadComments" and "getMaxScoreCheckers" they way too similar.

server/src/services/check.service.ts
```diff
@@ -0,0 +1,95 @@
...
+        .select('checkers."checkerId"')
+        .from(qb => {
+          return qb
+            .select('ts."checkerId"')
+            .addSelect('COUNT(score)', 'count')
+            .addSelect('AVG(score)', 'avg')
+            .from(TaskSolutionResult, 'ts')
+            .groupBy('ts."checkerId"')
+            .addGroupBy('ts."score"')
+            .having(`COUNT(score) = ${TASK_REVIEW_COUNT}`);
```
mikhama, line 47: Never do like that. It can leed to SQL injections in the future.
```
.having('COUNT(score) = :count', { count: TASK_REVIEW_COUNT }); 
```
but check the syntax it's probably just
```
.having('COUNT(score) = $1', [TASK_REVIEW_COUNT]); 
```
I'm not sure...

server/src/services/check.service.ts
```diff
@@ -0,0 +1,95 @@
...
+        }, 'checkers')
+        .groupBy('checkers."checkerId"')
+        .having('COUNT(checkers."checkerId") = 1')
+        .getQuery();
+      return `ts."checkerId" in ${subQuery}`;
+    })
+    .andWhere(
+      `ts."score" NOT BETWEEN "studentAvg"."avg" * ${LOW_ERROR_RATE} AND "studentAvg"."avg" * ${HIGH_ERROR_RATE}`,
```
mikhama, line 55: And please check all the places...

client/src/components/BadReview/BadReviewControllers.tsx
KostyaDedulya: +

client/src/components/BadReview/BadReviewControllers.tsx
KostyaDedulya: +

client/src/services/check.ts
KostyaDedulya: +

server/src/services/check.service.ts
KostyaDedulya: I don't know why, but in the first time with parametrs object i had an error(
Know i didn't have error
Fixed

server/src/services/check.service.ts
KostyaDedulya: I wouldn't refuse a call, because in this table pairs(checker, student) must be unique

I have made the query shorter

mikhama (approved):

KostyaDedulya:
#848

Issue #838, opened 2021-08-12
Schedule -improvements & fixes


not-SAINT:
**Improvements**
1. Add new feature - splited table rows by week
![image](https://user-images.githubusercontent.com/60464016/129205515-fc95b705-0d45-482a-b385-e2dd0f5bebe3.png)
![image](https://user-images.githubusercontent.com/60464016/129207865-1e562d95-2a01-4e0a-bbda-dfdf5611e6cc.png)
2. Add dates no List view 
![image](https://user-images.githubusercontent.com/60464016/129208998-7a5cf281-dfcf-4511-a96a-1b73c5e4d977.png)
3. Add new column - score 
![image](https://user-images.githubusercontent.com/60464016/129390939-844d47c4-afb8-4e28-b0e9-5f825463ab9b.png)

**UI fixes**
1. Fix Table columns width - evener columns width distribution
2. Improvements List view - add added margins
3. Fix icon style for color picker 
![image](https://user-images.githubusercontent.com/60464016/129208822-00426cc9-4568-411b-b881-0b39194214bc.png)
4. Calendar view - change first day of week to Monday

not-SAINT:
done.

PR #839, opened 2021-08-12
feat: schedule improvements (#838)

not-SAINT:
#838


client/src/styles/main.css
```diff
@@ -59,3 +59,11 @@ thead.ant-table-thead > tr > th {
 /* .rc-virtual-list-scrollbar {
   display: block !important;
 } */
+
+.table-row-current {
```
apalchys, line 63: Please add comment where it's used and why we need it

apalchys (approved):

PR #878, opened 2021-09-13
Fix/fix checks

KostyaDedulya:
1) remove button from modal
2) remove the second modal from bad comment
3) fix problem with number round
4) fix problems with guard


valerydluski (approved):

client/src/services/check.ts
```diff
@@ -18,15 +18,15 @@ export class CheckService {
...
     let dataFromService: IBadReview[] = [];
     try {
       switch (type) {
         case 'Bad comment':
-          dataFromService = await this.getDataFromServer(taskId, type);
+          dataFromService = await this.getDataFromServer(taskId, type, courseId);
           break;
         case 'Did not check':
-          dataFromService = await this.getDataFromServer(taskId, type);
+          dataFromService = await this.getDataFromServer(taskId, type, courseId);
```
mikhama, line 29: if this is the same request you can use shorten form:
```
case 'Bad comment':
case 'Did not check':
   dataFromService = await this.getDataFromServer(taskId, type, courseId);
   break;
```

mikhama: also you can return directly:
```
case 'something':
  return this.getDataFromServer(taskId, type, courseId);
default:
  return Promise.resolve([]);
```

mikhama (approved):

Issue #877, opened 2021-09-13
Task page has cross


Taneros:
**Describe the bug**

**To Reproduce**
1. Go to https://app.rs.school/course/schedule?course=js-fe-2021Q3 
2. Click on Click on any event
3. Scroll down to In the right corner there will be a cross 
4. See error: when you click on the cross "No Access
Probably you do not participate in the course. Please register or choose another course."

**Expected behavior**
Expect to not have that cross.

**Screenshots**

![](https://cdn.discordapp.com/attachments/604238535502856194/886675408790110238/unknown.png)

**Device (please complete the following information):**
 - OS: Windows 10
 - Browser Chrome
 - Version Version 1.29.79 Chromium: 93.0.4577.63 (Official Build) (64-bit)

**Additional context**

not-SAINT:
Fixed in #888 and may be closed.

PR #888, opened 2021-09-16
fix: restored alias for link (#877)

not-SAINT:
Fixed link for #877


apalchys (approved):

apalchys:
@not-SAINT Sorry for delay!

Issue #907, opened 2021-09-29
Feature Unassign Student doesn't work


splastiq:
**Describe the bug**
Feature Unassign Student doesn't work

**To Reproduce**
1. Go to main page of the app
2. Click on 'Expel/Unassign Student'
3. Click on 'Unassign'
4. Unassign doesn't work

**Expected behavior**
Feature Unassign Student should be worked correctly

**Screenshots**
![image](https://user-images.githubusercontent.com/18148907/135244823-1a4c1c6d-0809-4305-b7ef-15e684174430.png)

**Device (please complete the following information):**
 - OS: macOS
 - Browser chrome

PR #925, opened 2021-10-03
Fix unassign student bug

AlreadyBored:
Closes: #907 
Fix bug when mentor can't unassign student

### DIFF
--- a/server/src/routes/course/index.ts
+++ b/server/src/routes/course/index.ts
@@ -14,6 +14,7 @@ import {
   courseManagerGuard,
   basicAuthAws,
   anyCourseMentorGuard,
+  courseSupervisorOrMentorGuard,
 } from '../guards';
 import { setResponse } from '../utils';
 import {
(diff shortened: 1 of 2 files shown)


server/src/routes/course/student.ts
```diff
@@ -95,6 +95,20 @@ export const updateStudent = (_: ILogger) => async (ctx: Router.RouterContext) =
     setResponse(ctx, BAD_REQUEST, null);
     return;
   }
+  const user = ctx.state!.user;
+  const { githubId: userGithubId, isAdmin, isSupervisor, isManager } = user;
```
apalchys, line 99: it doesn't look right. `isSupervisor, isManager` are part of course roles/user

apalchys: try to reuse `userGuards` from `server/src/routes/guards.ts` 

```typescript
const guard = userGuards(ctx.state!.user);
const isPowerUser = guard.isPowerUser(courseId);
const isMentor = guard.isPowerUser(courseId);
...
```

AlreadyBored: done

apalchys (approved):

PR #954, opened 2021-10-20
Refactor and update design of CV page

apalchys:

```
 - Title should be a short informative summary of the pull request.
 - Do not put WIP in title. Use Draft PR (https://github.blog/2019-02-14-introducing-draft-pull-requests/)
```
 - [ ] I read and followed naming convention rules

---

#### 🤔 This is a ...
- [ ] New feature
- [ ] Bug fix
- [ ] Refactoring
- [ ] Test Case
- [ ] Documentation update
- [ ] Other

#### 🔗 Related issue link


AlreadyBored (changes requested):
It looks pretty cool in gerenal, just a few small remarks:
1. Contacts look good, but in the print version their icons disappear and it becomes unclear what these values mean.
This
contacts_1
Turns into this
contacts_2
2. Maybe we should make Self introduction video a clickable link at CV View?
3. Contacts with values that are close to their max length don't fit well in their container
contacts_3
4. There's no length limit on the Linkedin Profile input.
5. Do we really need text in top half of the CV in print version to be gray?
print_1
6. Old bug: the text on the buttons ceases to fit in them starting from the screen width of 586px and less. I suggested the solution above.
Before
buttons_1
After
buttons_2

PR #1186, opened 2022-02-03
feat(notifications): database schema

aaliakseyenka:

---

#### 🤔 This is a ...
- [ ] New feature
- [ ] Bug fix
- [ ] Refactoring
- [ ] Test Case
- [ ] Documentation update
- [ ] Other

#### 🔗 Related issue link


server/src/ormconfig.ts
```diff
@@ -10,7 +9,7 @@ const config: ConnectionOptions = {
   password: process.env.RSSHCOOL_API_PG_PASSWORD,
   database: process.env.RSSHCOOL_API_PG_DATABASE,
   entities: models,
-  migrations,
+  migrations: ['src/migrations/*.ts'],
```
aaliakseyenka, line 12: @apalchys is there any reason for having the `index` file? I removed it: nasty to update it all the times once you add migrations

apalchys: if i remember correctly, it doesn't work in production mode (container) with current setup because we don't have typescript and ts files inside container. but need to check

aaliakseyenka: understood. I have reverted for now. will check later and if ok, commit as separate change.

server/src/models/notification.ts
```diff
@@ -0,0 +1,37 @@
+import { Entity, Column, CreateDateColumn, PrimaryColumn, UpdateDateColumn, OneToMany, Index } from 'typeorm';
+import { NotificationChannelSettings } from './NotificationChannelSettings';
+
+export enum NotificationScope {
+  general = 'general',
```
apalchys, line 5: I've started to use PascalCase for enum

apalchys (approved):

PR #1205, opened 2022-02-11
feat: role guard to accept app and course roles

aaliakseyenka:
#### 🤔 This is a ...
- [x] Refactoring

Refactors guard to combine possible roles in single guard.


aaliakseyenka:
@apalchys 
"Merging can be performed automatically with 1 approving review." 😢

PR #1292, opened 2022-03-13
feat(notifications): tasks deadlines

aaliakseyenka:
#### 🤔 This is a ...

- [x] New feature
- [ ] Bug fix
- [ ] Refactoring
- [ ] Test Case
- [ ] Documentation update
- [ ] Other

Tracks tasks deadlines and notifies users


nestjs/src/courses/courseTasks/course-tasks.service.ts
```diff
@@ -62,4 +71,27 @@ export class CourseTasksService {
...
     });
   }
+
+  public getTasksPendingDeadline(
+    courseId: number,
+    { deadlineWithinHours = 24 }: { deadlineWithinHours?: number } = {},
+  ) {
+    const now = new Date().toISOString();
+    const endDate = new Date();
+    endDate.setHours(endDate.getHours() + deadlineWithinHours);
```
apalchys, line 81: i would probably suggest to bring some library for date management. If less experience developer will start contribute, it could be very hard to avoid mistakes etc.. :)

apalchys, line 81: i would probably suggest to bring some library for date management. If less experience developer will start contribute, it could be very hard to avoid mistakes etc.. :)

nestjs/src/courses/tasks/tasks.controller.ts
```diff
@@ -0,0 +1,35 @@
...
+
+@Controller('tasks')
+@ApiTags('courses tasks')
+@UseGuards(DefaultGuard, CourseGuard)
+export class TasksController {
+  private readonly logger = new Logger('tasks');
+
+  constructor(private tasksService: TasksService, private notificationService: NotificationsService) {}
+
+  @Post('/check-deadlines')
```
apalchys, line 16: similar to comment in the prev PR, maybe use `notify` word to make it implicit that it will notify

apalchys, line 16: similar to comment in the prev PR, maybe use `notify` word to make it implicit that it will notify

apalchys (approved):

nestjs/src/courses/courseTasks/course-tasks.service.ts
aaliakseyenka: added dayjs to handle dates

nestjs/src/courses/tasks/tasks.controller.ts
aaliakseyenka: thanks, updated

Issue #1179, opened 2022-02-01
Score: ability to highlight row 


dzmitry-varabei:
![image](https://user-images.githubusercontent.com/2429680/152068362-c2f6051d-017e-4342-aef7-ee1f97403cc9.png)

PR #1314, opened 2022-03-20
feat(score): ability to highlight row

ThorsAngerVaNeT:

---

#### 🤔 This is a ...

- [x] New feature
- [ ] Bug fix
- [ ] Refactoring
- [ ] Test Case
- [ ] Documentation update
- [ ] Other

#### 🔗 Related issue link

Implement #1179 

I have implemented 3 variations:
1. Radio button selection
2. Radio button and on row click selection
3. Hide radio buttons column and only on row click selection works

GIF 20 03 2022 17-00-55

### DIFF
--- a/client/src/modules/Score/components/ScoreTable/index.tsx
+++ b/client/src/modules/Score/components/ScoreTable/index.tsx
@@ -130,6 +130,8 @@ export function ScoreTable(props: Props) {
             setState(selectedRows.map(row => row.githubId));
           },
           type: 'radio',
+          columnWidth: 0,
+          renderCell: () => "",
         }}
         onRow={(record) => {
           return {
--- a/client/src/modules/Score/components/ScoreTable/index.tsx
+++ b/client/src/modules/Score/components/ScoreTable/index.tsx
@@ -131,6 +131,11 @@ export function ScoreTable(props: Props) {
           },
           type: 'radio',
         }}
+        onRow={(record) => {
+          return {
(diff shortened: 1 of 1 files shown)


apalchys:
@ThorsAngerVaNeT Thanks a lot for your contribution! Sorry, it took some time to review it. 

I'm really appreciate your effort but me and some our team members have concerns about this implementation. 
- Adding checkbox may add confusion since this UX pattern usually means that you select item for further actions
- It takes some space while Score is overloaded right now.

I understand that there is no built-it functionality in `antd` that allows to do it easy but I wonder what do you think about the following idea:
- When a user clicks row, save row index in a state
- Highlight row by adding a class if row index is in state.

What do you think about it?

ThorsAngerVaNeT:
Dear Andrei, thank you for your response and kind words.

> * Adding checkbox may add confusion since this UX pattern usually means that you select item for further actions
> * It takes some space while Score is overloaded right now.
> ...
> * When a user clicks row, save row index in a state
> * Highlight row by adding a class if row index is in state.
> 
> What do you think about it?

I totally agree and think the same way. That is why I implemented the third variant `Hide radio buttons column and only on row click selection works`. It's the current implementation in my branch, so if all requirements are 
GIF 28 03 2022 09-56-02

apalchys:
@ThorsAngerVaNeT well.. my bad. It seems I was confused and missed that the current implementation is the one that not in the gif. Thanks a lot!

Issue #1358, opened 2022-04-06
Schedule changes: notify course managers


aaliakseyenka:
When the course schedule is changed the notification should be sent to managers together with students

PR #1370, opened 2022-04-11
feat(notifications): notify mentors when schedule changes

aaliakseyenka:
#### 🤔 This is a ...

- [x] New feature
- [ ] Bug fix
- [ ] Refactoring
- [ ] Test Case
- [ ] Documentation update
- [ ] Other

#### 🔗 Related issue link
https://github.com/rolling-scopes/rsschool-app/issues/1358

when course schedule is changed, mentor will get notification as well


nestjs/src/schedule/schedule.service.ts
```diff
@@ -2,49 +2,65 @@ import { Injectable, Logger } from '@nestjs/common';
...
 
     for (const course of courses) {
-      const { students, ...rest } = course;
+      const { students, mentors, ...rest } = course;
+      const users = unionBy<{ userId: number }>(students, mentors, 'userId');
 
-      for (const student of students) {
-        const studentCourses = studentMap.get(student.userId) ?? [];
+      for (const user of users) {
+        const userCourses = recipientMap.get(user.userId) ?? [];
```
mikhama, line 57: I wondering can it be done faster by sql-request?

aaliakseyenka: what exactly? could you propose?

aaliakseyenka: seems i understood what you mean. will update

mikhama: I mean instead of all this operations with receiving students, mentors and users separately and putting it together with JS make something like this:
```sql
SELECT
  u.id,
  array_agg(coalesce(DISTINCT c."alias", '{}'::text)) as "courses"
FROM "user" u
LEFT JOIN student s ON u.id = s."userId"
LEFT JOIN mentor  m ON u.id = m."userId"
LEFT JOIN course  c ON c.id = s."courseId" OR c.id = m."courseId"
WHERE
  s."courseId" = $1 AND s."isExpelled" = false OR
  m."courseId" = $1 AND m."isExpelled" = false
GROUP BY u.id
```
And then just put it to the Map:
```javascript
const recipientMap = new Map<number, { alias: string }[]>(users.map(({ id, courses }) => [id, courses]);
```

aaliakseyenka: updated

mikhama (approved):

PR #1427, opened 2022-05-16
fix: handle discord rate limiter

agitaev:

* [X] I followed naming convention rules

---

#### 🤔 This is a ...

- [ ] New feature
- [X] Bug fix
- [ ] Refactoring
- [ ] Test Case
- [ ] Documentation update
- [ ] Other

#### 🔗 Related issue link

Issue #991 

Attempts to retry the request up to 5 times (can be tweaked) with 1000ms delay between requests, whenever it hits an error. Currently it retries on any errors, but can be configured to retry only on 429 errors.

- [X] Database migration is added or not needed
- [X] Documentation is updated/provided or not needed
- [X] Changes are tested locally

### DIFF
--- a/nestjs/src/gratitudes/discord.service.ts
+++ b/nestjs/src/gratitudes/discord.service.ts
@@ -1,6 +1,6 @@
 import { HttpService } from '@nestjs/axios';
 import { Injectable, Logger } from '@nestjs/common';
-import { lastValueFrom } from 'rxjs';
+import { map, retry } from 'rxjs';
 
 process.env.NODE_TLS_REJECT_UNAUTHORIZED = '0';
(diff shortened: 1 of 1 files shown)


apalchys:
If I get it right, you execute request but do not wait if it completed (no await).

Anyway, I think we need to do it differently.  If I remember correctly, we agreed at the meeting (sorry, it's not reflect in the issue) do not allow to send many gratitudes at once to avoid possible abuse. 

Let's do the following:
1. Create an endpoint that accepts array of gratitude messages
2. Fail if more than 5 gratitude messages
3. Limit to 5 recipients on UI side.

@dzmitry-varabei  @aaliakseyenka  @mikhama ?

Issue #1375, opened 2022-04-13
New "Your stats" widget


ktnktag:
**New "Your stats" widget**
Widget design changes.

Additional changes:
1. Remove the "status"
2. Add a line with the amount of% required to obtain a certificate

Figma [file ](https://www.figma.com/file/D3O7GJzjlG38YLh8FpmetI/RS-Design?node-id=121%3A11246)

**Additional context**
Since students do not understand what speedometers indicator shows, it serves as a decorative element. 
It also turned out that it is important for a student to see the current % - necessary to obtain a certificate
<img width="1135" alt="Frame 156" src="https://user-images.githubusercontent.com/101800066/163128744-d1f539b0-99b0-43ef-9ae5-1dbc397ebc0b.png">

I-vasilich-I:
Is it ready to implement or it's still on approval stage?

ktnktag:
Agreed with Dima, ready to implement

I-vasilich-I:
On position what total students count should be displayed: all or  only active? (80 / 5187)

ktnktag:
Your popsition / active students

ktnktag:
@I-vasilich-I in release missed info about necessary % to receive certificate

I-vasilich-I:
> @I-vasilich-I in release missed info about necessary % to receive certificate

@ktnktag I haven't missed it, @apalchys insisted that this info shouldn't be there:

> Sorry, I think it's better to remove right now. % and condition to get certificate are unique per course. There are no any formal rules right now and 70% could lead to confusion for example for ios courses or others.

I-vasilich-I:
@ktnktag how about adding icon in front of `Your stats` like on all other widgets?

![Screenshot from 2022-06-08 13-16-12](https://user-images.githubusercontent.com/67139199/172592771-faf56418-ee4e-4c40-a40d-d0e5c4f8a278.png)

ktnktag:
@I-vasilich-I Hello, 
We discussed it and decided to remove icons from all wigets 
If it not take a lot of time, can you remove them from others?

I-vasilich-I:
sure

I-vasilich-I:
@ktnktag I think this issue could be closed

PR #1515, opened 2022-06-13
fix: remove title icons from dashboard cards

I-vasilich-I:

---

#### 🤔 This is a ...

- [ ] New feature
- [ ] Bug fix
- [ ] Refactoring
- [ ] Test Case
- [ ] Documentation update
- [x] Other

#### 🔗 Related issue link

[issue#1375](https://github.com/rolling-scopes/rsschool-app/issues/1375)

Previously the title icon was removed on dashboard stats widget, this PR removes icons on all other dashboard cards for consistency as it was discussed on the related issue.
Screenshot from 2022-06-13 09-38-41

### DIFF
--- a/client/src/modules/StudentDashboard/components/MentorCard.tsx
+++ b/client/src/modules/StudentDashboard/components/MentorCard.tsx
@@ -1,7 +1,7 @@
 import * as React from 'react';
 import { Typography, List } from 'antd';
 import CommonCard from './CommonDashboardCard';
-import { TeamOutlined, GithubFilled, EnvironmentFilled } from '@ant-design/icons';
+import { GithubFilled, EnvironmentFilled } from '@ant-design/icons';
 import { MentorBasic } from 'common/models';
 import { GithubAvatar } from 'components/GithubAvatar';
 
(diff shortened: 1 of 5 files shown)


client/src/modules/StudentDashboard/components/MentorCard.tsx
```diff
@@ -60,7 +60,7 @@ export function MentorCard(props: Props) {
   return (
     <CommonCard
       title="Mentor"
-      icon={<TeamOutlined />}
+      icon={null}
```
aaliakseyenka, line 63: could we just omit the prop without specifying null?

I-vasilich-I: yes, I'll set icon prop type to be optional on component CommonCard and omit it on those cards. Seams like CommonCard is used only on this cards, so it wouldn't broke anything

I-vasilich-I: maybe I should remove that icon prop at all since it's not going to be used anywhere

aaliakseyenka (approved):

apalchys (approved):

Issue #1510, opened 2022-06-11
Score: Sorters on Cross-Check Table page are broken


AlreadyBored:
**Describe the bug**
On the Cross-Check Table page (`https://app.rs.school/course/admin/cross-check-table?course=some-course`) sorters by columns `Task`, `Checker` and `Student` don't work, clicking on them results in a request ending with an error (400 status code)  

**To Reproduce**

1. Go to Cross-Check Table page
2. Click on sorters of columns `Task`, `Checker` or `Student` 
3. See error

**Expected behavior**
Sorters work, no error occured

**Screenshots**
![cross-check-table-bug](https://user-images.githubusercontent.com/34455330/173195677-d2bdb53b-7446-4fff-9700-8897ee523b8b.JPG)

**Device (please complete the following information):**

- OS: Windows 10 Home
- Browser: Google Chrome
- Version: 102.0.5005.63

PR #1522, opened 2022-06-15
fix: cross-check sort and duplicate orderby param in api queries

MemoS77:

---

#### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Refactoring
- [ ] Test Case
- [ ] Documentation update
- [ ] Other

#### 🔗 Related issue link
#1510 

I discovered, when sorting by some fields and application make request to API, the orderBy parameter was duplicated. 
I propose the following solution - to check if it is an array, and if true, sort by the last selected field.

Screenshot_2

Now sorting in admin cross-check works. However, it doesn't hurt to check other places where the same API can be used.

Video_22-06-15_11-14-27(1)

### DIFF
--- a/client/src/api/api.ts
+++ b/client/src/api/api.ts
@@ -4581,7 +4581,7 @@ export const CoursesTasksApiAxiosParamCreator = function (configuration?: Config
             }
 
             if (orderBy !== undefined) {        
-                Array.isArray(orderBy)?localVarQueryParameter['orderBy'] = orderBy[0]:localVarQueryParameter['orderBy'] = orderBy;                
+                localVarQueryParameter['orderBy'] = Array.isArray(orderBy) ? orderBy[0]: orderBy;                
             }
 
             if (orderDirection !== undefined) {
--- a/client/src/api/api.ts
+++ b/client/src/api/api.ts
@@ -4580,8 +4580,8 @@ export const CoursesTasksApiAxiosParamCreator = function (configuration?: Config
                 localVarQueryParameter['current'] = current;
             }
 
-            if (orderBy !== undefined) {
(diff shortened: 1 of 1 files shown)


client/src/api/api.ts
```diff
@@ -4580,8 +4580,8 @@ export const CoursesTasksApiAxiosParamCreator = function (configuration?: Config
                 localVarQueryParameter['current'] = current;
             }
 
-            if (orderBy !== undefined) {
-                localVarQueryParameter['orderBy'] = orderBy;
+            if (orderBy !== undefined) {        
+                localVarQueryParameter['orderBy'] = Array.isArray(orderBy) ? orderBy[0]: orderBy;
```
apalchys, line 4584: please, do not modify this file manually. It's generated automatically and your changes will be lost when we update nestjs api and regenerate it.

apalchys: see above:

image

apalchys: Probably it makes to look into 

https://github.com/rolling-scopes/rsschool-app/blob/master/nestjs/src/courses/cross-checks/cross-checks.controller.ts#L26

or 

https://github.com/rolling-scopes/rsschool-app/blob/master/nestjs/src/courses/cross-checks/cross-checks.controller.ts#L38

MemoS77: > Probably it makes to look into

Понял. В бэк лезть не буду. Надо много времени, что бы в нем разобраться.  Закрываю PR. 
За этот issue взялся, так как подумал, что баг возник после моего добавление markdown на эту страницу, но оказалось, что это тут вообще не при чем, проблема в другом и она была раньше. Раз уж нашел причину и на локалке все заработало, отправил PR.

PR #1541, opened 2022-06-21
fix: handle onreset in markdown input

MemoS77:

#### 🤔 This is a ...

- [x] Bug fix

#### 🔗 Related issue link

#1513 

Reset text in textarea and preview of markdown component, switch to write mode if reset method of parent form handled.
Video_22-06-21_23-18-51

Component refactored, removed unused props. In application there is not cases of using other properties (besides notRequired) for the CommentInput component on the basis of which MarkdownInput was created and in place of which it can be applied.

### DIFF
--- a/client/src/components/Forms/MarkdownInput.tsx
+++ b/client/src/components/Forms/MarkdownInput.tsx
@@ -1,11 +1,14 @@
-import * as React from 'react';
+import { FC } from 'react';
 import { Input, Form, Button, Typography } from 'antd';
 import { useState } from 'react';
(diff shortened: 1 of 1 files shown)


apalchys (approved):

valerydluski (approved):

MemoS77:
👀🤔

apalchys:
@MemoS77 sorry, it seems you and @Pulya10c  persons worked on the same issue but @Pulya10c was merged despite that you did it first. my apologies. we will try to prevent such cases in the future.

Issue #1569, opened 2022-07-05
Re-distribution of Cross-Check doesn't work correctly


AlreadyBored:
**Describe the bug**
On trying to re-distribute course task with status "Completed" it's status is not changed to "Distributed", but response is Ok

**To Reproduce**

1. Go to course tasks management page
2. Click on "Complete" for cross-check task
3. Click on this "Distribute" for task mentioned above
4. See that task status remains "Completed" instead of "Distributed"

**Expected Behaviour**
Task status switches to "Distributed" after click on "Distribute" until next cron job tick

PR #1580, opened 2022-07-06
Fix: bug with broken re-distribution

AlreadyBored:

---

#### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Refactoring
- [ ] Test Case
- [ ] Documentation update
- [ ] Other

#### 🔗 Related issue link

https://github.com/rolling-scopes/rsschool-app/issues/1569

There's no sense in changing distribution mechanism, because it works well and doesn't need to have more complicated logic. Instead we should prevent user from making actions that lead to deadlock state.
Now Distribute and Complete buttons are disabled until submit deadline passed + Complete button can't be cliicked if task was not distributed


server/src/routes/course/crossCheck/isSubmissionDeadlinePassed.ts
```diff
@@ -0,0 +1,5 @@
+export const isSubmissionDeadlinePassed = (studentEndDate: string) => {
+  const currentTimestamp = new Date().getTime();
```
aaliakseyenka, line 2:
```suggestion
  const currentTimestamp = Date.now();
```

server/src/routes/course/crossCheck/createDistribution.ts
```diff
@@ -7,18 +7,25 @@ import { CrossCheckStatus } from '../../../models/courseTask';
...
   const { courseTaskId } = ctx.params;
 
   const courseTask = await taskService.getCourseTask(courseTaskId);
+
   if (courseTask == null) {
     setResponse(ctx, StatusCodes.BAD_REQUEST);
     return;
   }
 
+  if (!isSubmissionDeadlinePassed(courseTask.studentEndDate as string)) {
```
aaliakseyenka, line 24: why do you need that casting? you can efficiently pass both string and Date to Date constructor, in that way casting can be removed

server/src/routes/course/crossCheck/isSubmissionDeadlinePassed.ts
```diff
@@ -0,0 +1,5 @@
+export const isSubmissionDeadlinePassed = (studentEndDate: string) => {
```
aaliakseyenka, line 1: probably it is better to just put this function into tasks.service instead of creating new file per function

aaliakseyenka: in htis case it would be better to pass not date but task itself

client/src/pages/course/admin/tasks.tsx
```diff
@@ -162,6 +177,7 @@ function Page(props: CoursePageProps) {
       message.error('An error occurred.');
     } finally {
       setLoading(false);
+      await loadData();
```
aaliakseyenka, line 180: why to load data after removing spinner?

server/src/routes/course/crossCheck/isSubmissionDeadlinePassed.ts
AlreadyBored: Done

server/src/routes/course/crossCheck/createDistribution.ts
AlreadyBored: Done

server/src/routes/course/crossCheck/isSubmissionDeadlinePassed.ts
AlreadyBored: Done

client/src/pages/course/admin/tasks.tsx
AlreadyBored: Fixed

aaliakseyenka: not sure why you removed loading handling?
also probably you could just loaddata after mutating task.

client/src/pages/course/admin/tasks.tsx
```diff
@@ -145,23 +156,30 @@ function Page(props: CoursePageProps) {
...
+      } = await service.createCrossCheckDistribution(record.id);
+      if (crossCheckPairs.length) {
+        message.success('Cross-Check distrubtion has been created');
+      } else {
+        message.warn('Unable to create Cross-Check distribution: no solutions submitted');
+      }
     } catch (e) {
       message.error('An error occurred.');
+    } finally {
+      await loadData();
```
aaliakseyenka, line 170: same comment as below

AlreadyBored: loadData uses spinner by itself, and we are only interested in geting fresh data after request ends somehow, there's no need for 2 consecutive spinners
image

aaliakseyenka (approved):

PR #1532, opened 2022-06-20
fix: score table sorter

I-vasilich-I:
**🟢 Add `deploy` label if you want to deploy this Pull Request to staging environment**

#### 🧑‍⚖️ Pull Request Naming Convention

- Title should follow [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/#summary)
- Do not put issue id in title
- Do not put WIP in title. Use [Draft PR](https://github.blog/2019-02-14-introducing-draft-pull-requests/) functionality
- Consider to add `area:*` label(s)

* [x] I followed naming convention rules

---

#### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Performance optimization
- [ ] Refactoring
- [ ] Test Case
- [ ] Documentation update
- [ ] Other

#### 🔗 Related issue link

[issue#1526](https://github.com/rolling-scopes/rsschool-app/issues/1526)
[PR#1481](https://github.com/rolling-scopes/rsschool-app/pull/1481)

#### 💡 Background and solution

I returned `taskColumns: any[]` type and `sorter` as it was before my PR, which most likely will fix the issue, I can't be certain, because there're only 9 students on test db, no way to test if it works. 

I still don't get it why sorter works with just a column dataIndex/key and why provided sorter will sort only within current pagination section

UPDATE: set `taskColumns: Record<string, any>[]`

#### ☑️ Self Check before Merge

⚠️ Please check all items below before review. ⚠️

- [ ] Database migration is added or not needed
- [ ] Documentation is updated/provided or not needed
- [ ] Changes are tested locally

Issue #1526, opened 2022-06-16
Неправильная сортировка таблицы Score


natein:
Сортирует каждую страницу отдельно. Например сортируем по колонке Name:

Страница 1
![1](https://user-images.githubusercontent.com/12632270/174185186-803d6122-3d27-4807-b0bd-8df257ede855.PNG)

Страница 2
![2](https://user-images.githubusercontent.com/12632270/174185217-155a1ff0-743c-4e13-9ddd-90a530180def.PNG)

I-vasilich-I:
I'm pretty sure my sorter function from [PR#1481](https://github.com/rolling-scopes/rsschool-app/pull/1481) had broken it, I was afraid something like that could happen. I'll try to figure it out

P.S. Something wrong with # Column too (5803 id repeats a lot), looks like the same issue like [that one](https://github.com/rolling-scopes/rsschool-app/issues/1493)

I-vasilich-I:
issue fixed and could be closed

dzmitry-varabei:
@I-vasilich-I thank you!!

PR #1645, opened 2022-08-14
fix: cross-check table sorters

I-vasilich-I:

---

#### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Refactoring
- [ ] Test Case
- [ ] Documentation update
- [ ] Other

#### 🔗 Related issue link

[issue#1510](https://github.com/rolling-scopes/rsschool-app/issues/1510)

As was mentioned in [this PR](https://github.com/rolling-scopes/rsschool-app/pull/1522) problem caused by `orderBy` been an array, therefore request contains two orderBy queries and backend expects single value, so I figured it must be a client side problem and fixed it there. 

### DIFF
--- a/client/src/pages/course/admin/cross-check-table.tsx
+++ b/client/src/pages/course/admin/cross-check-table.tsx
@@ -9,10 +9,10 @@ import withCourseData from 'components/withCourseData';
 import { withSession } from 'components/withSession';
 import { isArray, omit } from 'lodash';
 import { useCallback, useEffect, useMemo, useState } from 'react';
-import { CourseService, CourseTaskDetails, CrossCheckPairs } from 'services/course';
+import { CourseService, CourseTaskDetails } from 'services/course';
 import { CoursePageProps } from 'services/models';
 import css from 'styled-jsx/css';
-import { CoursesTasksApi } from 'api';
+import { CoursesTasksApi, CrossCheckPairDto } from 'api';
 import PreparedComment from 'components/Forms/PreparedComment';
 
 const { Text } = Typography;
@@ -57,7 +57,7 @@ export function Page(props: CoursePageProps) {
   const [loading, setLoading] = useState(false);
   const [courseTasks, setCourseTasks] = useState<CourseTaskDetails[]>([]);
   const [crossCheckList, setCrossCheckList] = useState({
-    content: [] as CrossCheckPairs[],
+    content: [] as CrossCheckPairDto[],
     pagination: { current: 1, pageSize: 100 } as IPaginationInfo,
     orderBy: { field: DEFAULT_ORDER_BY, order: DEFAULT_ORDER_DIRECTION },
   });
@@ -92,7 +92,7 @@ export function Page(props: CoursePageProps) {
     async (
       pagination: TablePaginationConfig,
       filters: Record<keyof Filters, FilterValue | null>,
-      sorter: Sorter<CrossCheckPairs>,
+      sorter: Sorter<CrossCheckPairDto>,
     ) => {
       if (isArray(sorter)) {
         return;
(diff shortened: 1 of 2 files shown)


client/src/pages/course/admin/cross-check-table.tsx
```diff
@@ -86,8 +86,9 @@ export function Page(props: CoursePageProps) {
 
   const getCourseScore = useCallback(
     async (pagination: IPaginationInfo, filters: typeof fields, order: ScoreOrder) => {
+      const orderField = Array.isArray(order.field) ? order.field[0] : order.field;
```
ShekoMaxim, line 89: not good way create new crutch, if we don't fix reason why this situation happened
can you try delete `as any` in this row  ` onChange={handleChange as any}` number row is `173` and fixed problem with types.

I-vasilich-I: as far as I can tell `order.fields` comes from `dataIndex`, if it's an array so are fields, if it's not - fields not too.  We can use either this "crutch" or get rid of array in `dataIndex`, by mapping incoming data from `api.getCrossCheckPairs` or even better - changing that service on backend side, I've tried mapping incoming data + changing `dataIndex`, it works, but I don't think it's an improvement comparing to the first "crutch". 

When we use `dataIndex` as an array we instruct Table to look at nested props in data source. The `CrossCheckPairs` looks like this 
```ts
interface CrossCheckPairs {
  checker: {
    githubId: string;
    id: number;
  };
  task: {
    name: string;
    id: number;
  };
  student: {
    githubId: string;
    id: number;
  };
  url: string;
  comment: string;
  score: number;
  id: number;
}

```
therefore was used `dataIndex` like this: `["task", "name"]` but for ordering we only need `task`.
I mapped incoming data to 
```ts
interface CrossCheckPairs {
  checker: string;
  task: string
  student: string;
  url: string;
  comment: string;
  score: number;
  id: number;
}
```
and it works, but again it's probably not the solution we're looking for. 

So far this service is used only on this page, and I don't know if it was build only to be used here or it's going to be used somewhere else. 

So we can either check if `order.field` is an array, or change/add service to get data in more useful way, or maybe there's some other way to specify `order.field` which I'm not aware of, haven't found anything related on docs

ShekoMaxim: Why we do it in this case? 

> When we use dataIndex as an array we instruct Table to look at nested props in data source. 

If we use this service only on this page and for other tables we use ant sorting(without backend) in table, make sense use only `string` in `dataIndex` like `dataIndex: 'checker'` or `dataIndex: 'task'` in this case

I-vasilich-I: It was made this way 2 years ago. Anyway we can't just use `dataIndex` as a string without changing incoming data either on the client side or on the server side, because its main purpose is to access data to display it, sorting is like a side effect.

However I've just tried ant sorting `stringSorter('task.name')` and it seams like working, at least on the test db. 
So I'll leave `dataIndex` the way it is, adjust sorter and try to get rid of `any` type. 

Thanks for the tip.

I-vasilich-I: @ShekoMaxim Turns out using ant sorting isn't going to work here (same as on [Student Score table](https://github.com/rolling-scopes/rsschool-app/pull/1532)), because it works only with local data, it'll sort [within](https://github.com/rolling-scopes/rsschool-app/issues/1526) the current pagination - we'll get each pagination sorted separately, which is not what you'd expect - sorting the whole table. 

So we need to use sorting on the backend side which will return sorted and filtered data for current pagination. Therefore I fixed types, and added extra property for column type, where we can specify sorter field and use it on the request to the backend.

Any thoughts on this approach?

client/src/services/course.ts
```diff
@@ -780,4 +780,6 @@ export interface CrossCheckPairs {
   comment: string;
   score: number;
   id: number;
+  submittedDate: string;
```
apalchys, line 783: please remove this type and just use `CrossCheckPairDto` from `api`. It seems we don't need this type anymore

I-vasilich-I: done

apalchys (approved):

valerydluski (approved):

Issue #1793, opened 2022-10-18
Schedule: Buttons


ktnktag:
**Move buttons up to tabs**
Move buttons to the same row as tabs. 
![Group 500](https://user-images.githubusercontent.com/101800066/196546927-97dea830-f1ff-45e9-91ab-a2edf3c653d2.png)

Irregular actions (iCal Link, export, copy from) hide in the More button.
<img width="351" alt="Screenshot 2022-10-18 at 23 29 37" src="https://user-images.githubusercontent.com/101800066/196548547-9ee27d26-975c-40e7-99f0-6819e187868e.png">
[Figma file](https://www.figma.com/file/D31S4jMK4EQfnpvguVjc8W/RS-Redesign_Ant-design?node-id=130591%3A397007)

PR #1796, opened 2022-10-20
feat(schedule): move setting buttons under dropdown

nataliagulko:

---

#### 🤔 This is a ...

- [x] New feature
- [ ] Bug fix
- [x] Refactoring
- [ ] Test Case
- [ ] Documentation update
- [ ] Other

#### 🔗 Related issue link

#1793 

- iCal link, Export and Copy from buttons were moved under dropdown
- tests were added for new functionality

image


client/src/modules/Schedule/components/AdditionalActions/AdditionalActions.tsx
```diff
@@ -0,0 +1,90 @@
...
+        onCopyFromCourse();
+        break;
+      default:
+        break;
+    }
+  };
+
+  const dropdownMenu = <Menu items={menuItems} onClick={handleMenuItemClick} />;
+
+  return menuItems?.length !== 0 ? (
```
gorr2012, line 78: In this case and bellow should better use ```menuItems?.length !== 0 && <Dropdown .../>``` cause null we also render. 
https://reactjs.org/docs/conditional-rendering.html#preventing-component-from-rendering

nataliagulko: Are you sure?
After fix in SettingsPanel component error is shown:

> 'AdditionalActions' cannot be used as a JSX component.
>   Its return type 'false | Element' is not a valid JSX element.

gorr2012: Discussed

gorr2012 (approved):

apalchys (approved):

Issue #1914, opened 2022-12-21
score is not updated from 0, when students edited their reviewer comments on cross-check


adelinyshka:
**Describe the bug**
On cross-check, when student push button "amend comment" - student could change comment, but the score is not uploaded automatically.
This behaviour produce 0 score for many studens and force students who r making review to go to cross-check review again and update the score. 

**To Reproduce**

Go to cross-check review.
Add score and mark and push "Submit review as ..."
Find "History" on the page.
Push "Amend comment" button under the student reviewer comment.
You will see that comment is appeared in left field where first review was done. But no score, just 0 score

**Expected behavior**

Go to cross-check review.
Add score and mark and push "Submit review as ..."
Find "History" on the page.
Push **"Amend"** button under the student reviewer comment.
The comment **and score** are appeared in left field.

**Screenshots**
![image](https://user-images.githubusercontent.com/35169516/208936862-86c27ab8-645b-41da-b501-2994884098a7.png)

**Device (please complete the following information):**

- OS: [e.g. iOS] windows 11
- Browser [e.g. chrome, safari] chrome v.108.0.5359.125

**Additional context**
no

Issue #1894, opened 2022-12-11
Cross-check: Don't reset score in the cross check 


zzhanna:
**Course**
Node.js 2022 q4

**What is wrong**
My suggestion. For the new functionality of correcting the mark or message on the cross check. When I corrected myself today, I noticed that when you click edit, the score set is reset to zero .... And as a result, you can send 0 to the student for the work. I suggest not to reset the score as well as the message.
Today, I also got 206 at first, and the reviewer decided to correct the message, but did not pay attention to the score

PR #1921, opened 2022-12-23
fix(cross-check): change initial score

NikitaKorevo:

---

#### 🤔 This is a ...

- [x] New feature
- [x] Bug fix
- [ ] Refactoring
- [ ] Test Case
- [ ] Documentation update
- [ ] Other

#### 🔗 Related issue link

#1914
#1894

- Если нет прошлой оценки, в поле формы `score` будет `undefined` вместо нуля
- В поле формы `Comment` загружается последний отправленный комментарий

Before: 
before-score

After: 
after-score


client/src/components/CrossCheck/CrossCheckCriteriaForm.tsx
```diff
@@ -40,44 +40,44 @@ export interface CriteriaFormProps {
...
-      setTotalPoints(finalPoints > 0 ? finalPoints : 0);
-      return finalPoints > 0 ? finalPoints : 0;
-    }, [countStar, penalty]);
-  }
+  useEffect(() => {
+    form.setFieldValue('score', totalPoints);
+  }, [totalPoints]);
+
+  useEffect(() => {
+    if (criteriaData.length === 0) return;
```
AlreadyBored, line 69: Minor suggestion: this is shorter and more common
```suggestion
   if (!criteriaData.length) return;
```

client/src/components/CrossCheck/CrossCheckCriteriaForm.tsx
```diff
@@ -40,44 +40,44 @@ export interface CriteriaFormProps {
...
-  const calculationResultPoints = calculateTotalPoints();
+    const sumPoints = +countStar
+      .map(item => item.point)
+      .reduce((prev, next) => prev + next, 0)
+      .toFixed(1);
+    const sumPenalty = penalty.map(item => item.point).reduce((prev, next) => prev + next, 0);
+    let finalPoints = sumPoints + sumPenalty;
+    finalPoints = finalPoints > 0 ? finalPoints : 0;
+
+    form.setFieldValue('score', finalPoints);
```
AlreadyBored, line 79: Not 100% related to your changes, but still:
1. We could remove excessive `map` iteration, `reduce` is enough here
2. We could avoid usage of `let` here, it's not necessary :)
I would change it to something like this
```suggestion
    const scorePoints = +countStar
      .reduce((acc, next) => acc + next.point, 0)
      .toFixed(1);

    const penaltyPoints = penalty.reduce((acc, next) => acc + next.point, 0);

    const totalScore = scorePoints + penaltyPoints;
    const finalScore = totalScore > 0 ? totalScore : 0;

    form.setFieldValue('score', finalScore);
```

client/src/pages/course/student/cross-check-review.tsx
```diff
@@ -53,16 +53,21 @@ function Page(props: CoursePageProps) {
...
   const { value: courseTasks = [] } = useAsync(() => courseService.getCourseCrossCheckTasks(), [props.course.id]);
 
   const loadStudentScoreHistory = async (githubId: string) => {
-    const taskCriteriaData = await taskServise.getCriteriaForCourseTask(criteriaId as number);
+    setState({ loading: true, data: [] });
+
+    const [taskCriteriaData, result] = await Promise.all([
+      taskService.getCriteriaForCourseTask(criteriaId as number),
+      courseService.getTaskSolutionResult(githubId, courseTaskId as number),
+    ]);
```
AlreadyBored, line 66: We should avoid typecasts as much as possible. It seems that we should have both `taskCriteriaData` and `courseTaskId` to make proper requests, so it's better to have early return if any of these values is `null` instead of typecast here

client/src/pages/course/student/cross-check-review.tsx
```diff
@@ -89,10 +94,13 @@ function Page(props: CoursePageProps) {
       };
     });
 
-    setState({ loading: false, data: solutionReviews ?? [] });
+    const activeSolutionReview = solutionReviews[0];
+
+    form.setFieldValue('comment', activeSolutionReview.comment.slice(markdownLabel.length));
     if (result !== null) {
```
AlreadyBored, line 100: It seems that this check here is not necessary, we already checked that `result !== null` on line 72, could you please remove this `if` statement?

client/src/pages/course/student/cross-check-review.tsx
```diff
@@ -89,10 +94,13 @@ function Page(props: CoursePageProps) {
       };
     });
 
-    setState({ loading: false, data: solutionReviews ?? [] });
+    const activeSolutionReview = solutionReviews[0];
```
valerydluski, line 97: I think it would be better to use destructuring here.

andreikovna:
In my opinion button Amend comment loses its sense. 
If we have review history comment will appear automatically in the form with this fix. What do you think about it? Is it necessary to leave this functionality or this button can be removed?

NikitaKorevo:
> In my opinion button Amend comment loses its sense. If we have review history comment will appear automatically in the form with this fix. What do you think about it? Is it necessary to leave this functionality or this button can be removed?

Я могу удалить ее в следующем PR

valerydluski (approved):

AlreadyBored (approved):

PR #2037, opened 2023-02-24
Cross check criteria validation while adding task 

andreikovna:

---

#### 🤔 This is a ...

- [x] New feature
- [ ] Bug fix
- [ ] Refactoring
- [ ] Test Case
- [ ] Documentation update
- [ ] Other

#### 🔗 Related issue link

Part of #664 

image

1. Disabled button Add criteria without description and points for criteria
image

2. Warning about lack of criteria if  download JSON without criteria
image

3. Highlighting the line in red with hint if after downloading JSON points for criteria equal null
image

4. Disabled submit adding/updating task if any criteria point has zero value
image


client/src/modules/CrossCheck/__tests__/AddCriteriaForCrossCheck.test.tsx
```diff
@@ -19,6 +19,14 @@ describe('AddCriteriaForCrossCheck', () => {
 
   test('should call addCriteria when "Add new criteria" button was clicked', () => {
     render(<AddCriteriaForCrossCheck onCreate={addCriteria} />);
+    const selectCriteriaType = screen.getByRole('combobox') as HTMLSelectElement;
+    fireEvent.mouseDown(selectCriteriaType);
+    const optionTitle = screen.getByTestId('Title') as HTMLOptionElement;
```
AlreadyBored, line 24: Queries are generics that accept returned element type. So if you really want to provide returned element type, it's better to do it that way to avoid typecast:  
  ```suggestion
      const selectCriteriaType = screen.getByRole<HTMLSelectElement>('combobox');
      fireEvent.mouseDown(selectCriteriaType);
      const optionTitle = screen.getByTestId<HTMLOptionElement>('Title');
```
But in this particular case I am not sure that we need it here at all :)

File: client/src/pages/admin/tasks.tsx [hunk 2]
```diff
@@ -63,6 +64,21 @@ function Page(props: Props) {
...
   const handleModalSubmit = useCallback(
     async (values: any) => {
+      const checkCriteria = () => {
+        return dataCriteria.every(item => {
+          if (item.type !== TaskType.Title) {
+            return item.max !== 0;
+          }
+          return;
+        });
+      };
```
AlreadyBored, line 74: 1. `return` on line 72 is unnecessary, we could safely remove it
2. It seems that we can add penalty with positive score, please consider to add one more check for this case

client/src/pages/admin/tasks.tsx
```diff
@@ -63,6 +64,21 @@ function Page(props: Props) {
...
+
+      const isVerified = dataCriteria.length ? checkCriteria() : true;
```
AlreadyBored, line 76: Minor: I would shortened it a bit  
```suggestion
      const isVerified = !dataCriteria.length || checkCriteria();
```

apalchys: i think i disagree with the suggested solution as it is harder to understand the condition. the first is clear and simple

client/src/modules/CrossCheck/EditableCellForCrossCheck.tsx
```diff
@@ -1,42 +1,44 @@
...
+    ) : type !== TaskType.Title ? (
+      <InputNumber style={{ width: 65 }} />
+    ) : null;
 
   return (
-    <td {...restProps}>
-      {editing ? (
+    <td
+      {...restProps}
+      title={points === 0 ? 'Check points for this line' : ''}
```
apalchys, line 38: minor, but i would extract `points === 0` into a variable

apalchys (approved):

client/src/pages/admin/tasks.tsx
andreikovna: For the first note - actually I missed `true`  :(
For the second - for these cases to avoid positive penalties we use only the absolute value of a number for following calculating the final score

client/src/modules/CrossCheck/EditableCellForCrossCheck.tsx
andreikovna: Thanks for the review!
Fixed it

client/src/modules/CrossCheck/__tests__/AddCriteriaForCrossCheck.test.tsx
andreikovna: Maksim, thanks a lot for review! I've made necessary changes

valerydluski (approved):

AlreadyBored (approved):
Thank you!

Issue #1889, opened 2022-12-08
Score: Add trim to whitespaces to search fields on score page


aibolit666:
**Describe the feature**
When you paste from clipboard into the search field the name of a student or a nickname of a mentor, it happens that when copying, spaces are captured at the beginning or at the end of the value. After sending such a value with extra spaces, the search is not performed. I suggest that before sending the value to the search field, trim the spaces at the beginning and at the end.

**Additional context**
I think the trim() method is great for this feature.

PR #2137, opened 2023-05-03
fix:  Add trim to whitespaces to search fields on score page

konstantinplayneta:

---

#### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Refactoring
- [ ] Test Case
- [ ] Documentation update
- [ ] Other

#### 🔗 Related issue link

https://github.com/rolling-scopes/rsschool-app/issues/1889

Before
image

After
image

### DIFF
--- a/client/src/components/UserSearch.tsx
+++ b/client/src/components/UserSearch.tsx
@@ -32,6 +32,7 @@ export function UserSearch(props: UserProps) {
 
   const handleSearch = async (value: string) => {
     if (value) {
+      value = value.trim();
       const data = await searchFn(value, onlyStudentsWithoutMentorShown);
       setData(data);
     } else {
--- a/client/src/components/UserSearch.tsx
+++ b/client/src/components/UserSearch.tsx
@@ -31,8 +31,8 @@ export function UserSearch(props: UserProps) {
   }, [props.defaultValues]);
 
   const handleSearch = async (value: string) => {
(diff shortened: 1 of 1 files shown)


client/src/components/UserSearch.tsx
```diff
@@ -32,6 +32,7 @@ export function UserSearch(props: UserProps) {
 
   const handleSearch = async (value: string) => {
     if (value) {
+      value = value.trim();
```
aaliakseyenka, line 35: I suppose it makes sense to validate trimmed value inside `if`, ie do trim beforehand

konstantinplayneta: I did not understand at what point it is necessary to make a trim? Do you mean before the function (onSearch={val => handleSearch(val.trim())}) or before the if block? Something like this?                             
```
value = value.trim();                                       
 if (value) {
   const data = await searchFn(value, onlyStudentsWithoutMentorShown);
   ...
}
```

aaliakseyenka: like in your example above. the thing is that if value=` `, it will pass the `if` condition, but due to trim you will call search with empty value

konstantinplayneta: fix this

aaliakseyenka (approved):

PR #2250, opened 2023-08-07
Fix: users search by fullname, add debounce to user search

ThorsAngerVaNeT:

---

#### 🤔 This is a ...

- [x] New feature
- [x] Bug fix
- [ ] Refactoring
- [ ] Test Case
- [ ] Documentation update
- [ ] Other

#### 🔗 Related issue link

1) Fix the impossible to search by user's full name on the Gratitude page
2) Add debounce for Person input to prevent unnecessary API calls

Before
chrome_U35oTq3zaf

After
chrome_6wbZxzAdsS


client/src/components/UserSearch.tsx
```diff
@@ -49,7 +49,7 @@ export function UserSearch(props: UserProps) {
       defaultActiveFirstOption={false}
       showArrow={defaultValues ? Boolean(defaultValues.length) : false}
       filterOption={false}
-      onSearch={handleSearch}
+      onSearch={debounce(handleSearch, 300)}
```
valerydluski, line 52: I think it's better to use useDebounce
https://github.com/streamich/react-use/blob/master/docs/useDebounce.md

ThorsAngerVaNeT: > I think it's better to use useDebounce https://github.com/streamich/react-use/blob/master/docs/useDebounce.md

Like this?
```ts
  const [data, setData] = useState<Person[]>([]);
  const [value, setValue] = useState<string>('');
  const [debouncedValue, setDebouncedValue] = useState<string>('');
  const {
    searchFn = defaultSearch,
    defaultValues,
    keyField,
    showMentor,
    onlyStudentsWithoutMentorShown,
    ...otherProps
  } = props;

  useEffect(() => {
    setData(defaultValues ?? []);
  }, [props.defaultValues]);

  useEffect(() => {
    if (debouncedValue) {
      searchFn(debouncedValue, onlyStudentsWithoutMentorShown).then(data => setData(data));
    } else {
      setData(props.defaultValues ?? []);
    }
  }, [debouncedValue]);

  useDebounce(
    () => {
      setDebouncedValue(value);
    },
    300,
    [value],
  );

  const handleSearch = (value: string) => setValue(value.trim());
```

ThorsAngerVaNeT: > yes

I have pushed updated solution

apalchys: @ThorsAngerVaNeT it could be done much simpler using `lodash/debounce`

```typescript
import debounce from 'lodash/debounce';
// ...

const handleSearch = debounce(async (value: string) => {
  value = value.trim();
 ...
}, 1000);

ThorsAngerVaNeT: > @ThorsAngerVaNeT it could be done much simpler using `lodash/debounce`
> 
> ```ts
> import debounce from 'lodash/debounce';
> // ...
> 
> const handleSearch = debounce(async (value: string) => {
>   value = value.trim();
>  ...
> }, 1000);
> ```

You are right. I did it the same way in the beginning, but I used debounce in setting the onSearch prop at line 52. I like the way you did it.

valerydluski: You can ignore my comment

ThorsAngerVaNeT: > You can ignore my comment

Got it :)

valerydluski (approved):

PR #2274, opened 2023-08-20
feat: add navigation menu

valerydluski:

---

#### 🤔 This is a ...

- [ ] New feature
- [ ] Bug fix
- [ ] Refactoring
- [ ] Test Case
- [ ] Documentation update
- [ ] Other

#### 🔗 Related issue link

image
image


client/src/modules/Course/contexts/SessionContext.tsx
```diff
@@ -18,7 +23,7 @@ type Props = React.PropsWithChildren<{
   adminOnly?: boolean;
 }>;
 
-export function SessionProvider(props: Props) {
+export function DefaultPageProvider(props: Props) {
```
aaliakseyenka, line 26: I would still name it SessionProvider

client/src/modules/Course/contexts/SessionContext.tsx
```diff
@@ -8,7 +8,12 @@ import Router from 'next/router';
 import React, { useEffect } from 'react';
 import { useAsync } from 'react-use';
 
-export const SessionContext = React.createContext<Session>({} as Session);
+type DefaultContext = {
+  session: Session;
+  activeCourse?: ProfileCourseDto;
```
aaliakseyenka, line 13: doesn't it make sense to separate. ie Session and CourseContext are different to me. Not all pages have course, so I would split to Session & Course. Plus we have this `withCourseData`, which could be eventually transformed to context provider as well. in this case you won't need to mix up these things

client/src/pages/404.tsx
```diff
@@ -23,4 +20,12 @@ function Page(props: IProps) {
   );
 }
 
-export default withSession(Page);
+function PageWithContext() {
+  return (
+    <DefaultPageProvider>
```
aaliakseyenka, line 25: consider inlining into the `Page`

Issue #1766, opened 2022-09-28
[Schedule] add filter for status column in all tab


valerydluski:
![image](https://user-images.githubusercontent.com/57412523/192847915-0cce91dd-a557-44d4-84e0-aaf48af1c638.png)
Add the ability to filter by statuses on the all tab.
multiple statuses can be selected.
status filters should not be applied on other tabs

PR #2284, opened 2023-09-06
feat(schedule): add status column filter

ThorsAngerVaNeT:

---

#### 🤔 This is a ...

- [x] New feature
- [ ] Bug fix
- [ ] Refactoring
- [ ] Test Case
- [ ] Documentation update
- [ ] Other

#### 🔗 Related issue link

https://github.com/rolling-scopes/rsschool-app/issues/1766

The `status` column filter is reset when a tab is changed from the `All` tab.
image


client/src/modules/Schedule/components/TableView/TableView.tsx
```diff
@@ -122,39 +135,99 @@ export interface TableViewProps {
...
+        break;
+      case ColumnName.Status:
+        setCombinedFilter({
+          ...combinedFilter,
+          statuses: combinedFilter.statuses.filter(tag => tag !== removedTagValue.trim().toLowerCase()),
+          filterTags,
+        });
+        break;
+      default:
+        message.error('An error occurred. Please try again later.');
```
apalchys, line 209:
```suggestion
        message.error('Unknown tag');
```

client/src/modules/Schedule/components/TableView/TableView.tsx
```diff
@@ -122,39 +135,99 @@ export interface TableViewProps {
   statusFilter?: string;
 }
 
+export interface CombinedFilter {
```
apalchys, line 138: prefer use `type` in such cases

client/src/modules/Schedule/components/TableView/TableView.tsx
```diff
@@ -122,39 +135,99 @@ export interface TableViewProps {
...
+  types: string[];
+  statuses: string[];
+
+  filterTags?: string[];
```
apalchys, line 142: maybe just `tags` ?

client/src/modules/Schedule/components/TableView/TableView.tsx
```diff
@@ -122,39 +135,99 @@ export interface TableViewProps {
...
+  };
+
+  const handleTableChange = (_: any, filters: Record<ColumnKey, FilterValue | string[] | null>) => {
+    const combinedFilter: CombinedFilter = {
+      types: filters.type?.map(tag => tag.toString()) ?? [],
+      statuses: filters.status?.map(status => status.toString()) ?? [],
+    };
+
+    combinedFilter.filterTags = [
+      ...combinedFilter.types.map(tag => `${ColumnName.Type}: ${TAG_NAME_MAP[tag as CourseScheduleItemDto['tag']]}`),
```
apalchys, line 225: no need to use label as value and later split by ':' and using reverse map.
For example, consider using the following structure instead of just string:

```typescript
type FilterTag = {
  label: string
  value: string 
  tagType: 'type' | 'status'
}
```

client/src/modules/Schedule/components/TableView/TableView.tsx
ThorsAngerVaNeT: Sure, could you please explain why so I could avoid the same problems in the future?
I thought it's more of a personal preference

client/src/modules/Schedule/components/TableView/TableView.tsx
ThorsAngerVaNeT: fixed

client/src/modules/Schedule/components/TableView/TableView.tsx
ThorsAngerVaNeT: @apalchys could you please check if I understand and implement your suggestion correctly?
I was thinking about making `combinedFilter` as an array of `FilterTag` items, but that way I believe we are forced to filter that array all the time to separate items by type

client/src/modules/Schedule/components/TableView/TableView.tsx
ThorsAngerVaNeT: fixed

client/src/modules/Schedule/components/TableView/TableView.tsx
apalchys: for consistency. I believe we mostly use `type` for things which are not supposed to be extended like interfaces. If it's an internal thing inside component, I would use `type` due to more available features. (e.g. you can't combine `string | { foo: string}` via interface)

ThorsAngerVaNeT: > for consistency. I believe we mostly use `type` for things which are not supposed to be extended like interfaces. If it's an internal thing inside component, I would use `type` due to more available features. (e.g. you can't combine `string | { foo: string}` via interface)

Got it, thank you

client/src/modules/Schedule/components/TableView/TableView.tsx
apalchys: it's better imno. thanks.

apalchys (approved):

PR #2303, opened 2023-09-14
fix: find checker user id

aaliakseyenka:
#### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Refactoring
- [ ] Test Case
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/server/src/routes/course/crossCheck/createMessage.ts
+++ b/server/src/routes/course/crossCheck/createMessage.ts
@@ -1,7 +1,8 @@
 import Router from '@koa/router';
+import { getRepository } from 'typeorm';
 import { StatusCodes } from 'http-status-codes';
 import { ILogger } from '../../../logger';
-import { IUserSession } from '../../../models';
+import { IUserSession, Student } from '../../../models';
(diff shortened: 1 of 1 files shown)


server/src/routes/course/crossCheck/createMessage.ts
```diff
@@ -67,9 +68,15 @@ export const createMessage = (_: ILogger) => async (ctx: Router.RouterContext) =
     user: user,
   });
 
+  const userId = await getUserId(student.id, taskSolutionResult.checkerId, inputData.role);
+  if (!userId) {
+    setErrorResponse(ctx, StatusCodes.BAD_REQUEST, 'user not found');
```
I-vasilich-I, line 73: how about `NOT_FOUND` status code?

aaliakseyenka: IMO this is a bad request as we can't process current action due to missing data. It might be not found if we couldn't find cross check result for example.

valerydluski (approved):

Issue #1980, opened 2023-01-28
redesign and refactoring of the admin/task page.


valerydluski:

nataliagulko:
This page has huge performance issue. When I type something in fields or select items from dropdown in Modal, the whole page with table and all data is re-rendered. Need to be refactored into separate components and logical parts

TattiMikhailava:
There are some difficulties with filling in the form fields. The coordinators show some properties, but these options do not help to organize the table data. It is also not clear how and when to mark task settings. One of the required fill ‘Task type’ defines settings of the assignment, if they are necessary to be marked. 

![task add current](https://user-images.githubusercontent.com/100757852/222109349-bef0a616-60ae-4c5e-9224-fa2c81f140aa.png)

**Design solution** 

1. Add tips to fields. 

2. Organize the task list in alphabetical order: 
Codejam  
Codewars  
CV (HTML)  
CV (Markdown)  
Google form test  
Group task  
HTML task  
Interview  
JS task  
Jupyter Notebook  
Kotlin task  
Objective-C task  
RS School app test  
Technical screening  
Cross-check. 

4. Divide tasks into groups according to the required characteristics: ‘Criteria for the cross-check task’, ‘GitHub’, ‘JSON Attributes’.  

5. Show these criteria as available if they are needed to be filled for some of tasks. Otherwise, the collapse component must have a disabled state. 

The group of tasks, that has required settings for the cross-check: 

1.     Cross-check  
2.     Group task 

![cross-check](https://user-images.githubusercontent.com/100757852/222109815-573d3213-adbb-43e0-95c0-b417c25e9ef5.png)



The group of tasks, that has required settings for Github: 

1.     Codejam 
2.     JS task 

![GitHub](https://user-images.githubusercontent.com/100757852/222109902-1c86314f-bd39-4f28-ad1a-a107141f46e3.png)



The group of tasks, that has required settings for the JSON Attributes: 

1.     Interview  
2.     Technical screening  
3.     RS School app test  
4.     Codewars 

![JSON](https://user-images.githubusercontent.com/100757852/222110003-4cdfa984-7fd0-413e-a74b-75db250d09a3.png)



The group of tasks, that has required settings for ‘GitHub’ and ‘JSON Attributes’: 

1.     Jupyter Notebook 
2.     Kotlin task 
3.     Objective-C task 

[...]

PR #2305, opened 2023-09-21
refactor: admin tasks page

I-vasilich-I:

---

#### 🤔 This is a ...

- [ ] New feature
- [ ] Bug fix
- [x] Refactoring
- [x] Test Case
- [ ] Documentation update
- [ ] Other

#### 🔗 Related issue link

[issue#1980](https://github.com/rolling-scopes/rsschool-app/issues/1980#issuecomment-1407339366)

So far just extracted components and added tests, logic's untouched

SIDE CHANGES:
- Minor change on Course Task Modal: task type now get auto set from chosen task

- remove typeEnum and instead use TaskType enum from `@entities`, which leaves `TaskDtoTypeEnum === CourseTaskDtoTypeEnum` and removes possible double source data issue. Generated `api` - no changes.

QUESTIONS:
- is there any purpose to have `task type` on course task? 


client/src/modules/Tasks/pages/TasksPage/TasksPage.tsx
```diff
@@ -0,0 +1,146 @@
...
+    disciplineId: values.discipline,
+    attributes: JSON.parse(values.attributes ?? '{}') as Record<string, unknown>,
+  };
+  return data;
+}
+
+function prepareValues(modalData: TaskDto) {
+  if (!modalData) {
+    return modalData;
+  }
```
valerydluski, line 141: modal data is declared with the type taskDto, as I understand it, this if block is not needed

I-vasilich-I: I find it strange too, and for 99% sure it is not needed, but that 1%)) 
Considering it's just a refactoring I left it as it is, just in case...

valerydluski: Isn't that the point of refactoring, to make the code more understandable?))
If there can be null or undefined here, let's indicate it in the argument typing.

I-vasilich-I: I took another look at it, and it could be only a `TaskDto` )) 

removed it

apalchys:
@I-vasilich-I 

> is there any purpose to have task type on course task?

It was possible to override type, at least in the past and I think we need to preserve it for compatibility

client/src/modules/CourseManagement/components/CourseTaskModal/index.tsx
```diff
@@ -67,29 +68,38 @@ export function CourseTaskModal(props: Props) {
     props.onCancel();
   };
 
+  const findTaskById = useCallback((id: number | undefined) => tasks.find(t => t.id === id), [tasks]);
```
valerydluski, line 71:
```suggestion
  const findTaskById = useCallback((id: number) => tasks.find(t => t.id === id), [tasks]);
```

client/src/modules/CourseManagement/components/CourseTaskModal/index.tsx
```diff
@@ -67,29 +68,38 @@ export function CourseTaskModal(props: Props) {
...
+
   const filterOption = useCallback(
     (input: string, option?: { value: number }): boolean => {
       if (!input) {
```
valerydluski, line 75:
```suggestion
      if (!input || !option) {
```

client/src/modules/Tasks/components/CrossCheckTaskCriteriaPanel/CrossCheckTaskCriteriaPanel.tsx
```diff
@@ -0,0 +1,43 @@
...
+  UploadCriteriaJSON,
+  AddCriteriaForCrossCheck,
+  EditableTable,
+  ExportJSONButton,
+} from 'modules/CrossCheck';
+import { LABELS } from 'modules/Tasks/constants';
+
+type Props = {
+  dataCriteria: CriteriaDto[];
+  setDataCriteria: (c: CriteriaDto[]) => void;
```
valerydluski, line 14:
```suggestion
  setDataCriteria: (criteria: CriteriaDto[]) => void;
```

client/src/modules/Tasks/components/TableView/TableView.tsx
```diff
@@ -0,0 +1,121 @@
...
+    },
+  ];
+}
+
+type Props = {
+  data: TaskDto[];
+  handleEditItem: (record: TaskDto) => Promise<void>;
+};
+
+export const TableView = ({ data, handleEditItem }: Props) => {
```
valerydluski, line 99: I would rename it, because seeing `TableView` makes me expect something like `CalendarView`, for example. Probably something like `TasksTable` would be better.

client/src/modules/Tasks/components/TaskModal/TaskModal.tsx
```diff
@@ -0,0 +1,159 @@
...
+    () =>
+      union(
+        tasks
+          .map(task => task.skills || [])
+          .concat(SKILLS)
+          .flat()
+          .sort(),
+      ),
+    [tasks],
+  );
```
valerydluski, line 49: Try something like this:
```
const allSkills = useMemo(
() => union(tasks.flatMap(task => task.skills || []).concat(SKILLS).sort()),
[tasks],
)
```

client/src/modules/Tasks/components/TableView/TableView.tsx
I-vasilich-I: I think I took the naming from schedule table) Have no clue how `CalendarView` looks like) 
Nevertheless `TasksTable` makes sense, I'll rename it.

valerydluski (approved):

PR #2384, opened 2023-12-06
feat: implement feature trim function to the search inputs on the score

Zoro444:

---

#### 🤔 This is a ...

- [x] New feature
- [ ] Bug fix
- [ ] Refactoring
- [ ] Test Case
- [ ] Documentation update
- [ ] Other

#### 🔗 Related issue link
https://app.rs.school/course/score

### DIFF
--- a/client/src/modules/Score/components/ScoreTable/index.tsx
+++ b/client/src/modules/Score/components/ScoreTable/index.tsx
@@ -38,22 +38,6 @@ type StudentsState = {
 
 const courseTasksApi = new CoursesTasksApi();
 
-function trimFiltersProps(filters: ScoreTableFiltersModified): ScoreTableFiltersModified {
-  const trimmedFiltersProps = structuredClone(filters);
-  const filtersKeys = Object.keys(filters) as (keyof ScoreTableFiltersModified)[];
-
-  filtersKeys.forEach(key => {
-    const filterKey = filters[key];
-    if (filterKey && typeof filterKey !== 'boolean') {
-      key = key as Exclude<keyof ScoreTableFiltersModified, 'activeOnly'>;
-      if (Array.isArray(filterKey)) {
-        trimmedFiltersProps[key] = filterKey.map(elem => elem.trim());
-      }
-    }
-  });
-  return trimmedFiltersProps;
-}
-
 export function ScoreTable(props: Props) {
   const router = useRouter();
   const { width } = useWindowDimensions();
@@ -191,10 +175,9 @@ export function ScoreTable(props: Props) {
 
     try {
       props.onLoading(true);
-      const trimmedFiltersProps = trimFiltersProps(filters);
       const [studentCourseScore] = await Promise.all([
         courseService.getStudentCourseScore(props.session?.githubId as string),
-        getCourseScore(pagination, trimmedFiltersProps, sorter),
+        getCourseScore(pagination, filters, sorter),
       ]);
       setStudentScore(studentCourseScore);
     } finally {
--- a/client/src/modules/Score/components/ScoreTable/index.tsx
+++ b/client/src/modules/Score/components/ScoreTable/index.tsx
@@ -48,8 +48,6 @@ function trimFiltersProps(filters: ScoreTableFiltersModified): ScoreTableFilters
       key = key as Exclude<keyof ScoreTableFiltersModified, 'activeOnly'>;
       if (Array.isArray(filterKey)) {
         trimmedFiltersProps[key] = filterKey.map(elem => elem.trim());
-      } else {
-        trimmedFiltersProps[key] = filterKey.trim();
       }
     }
   });
--- a/client/src/modules/Score/components/ScoreTable/index.tsx
+++ b/client/src/modules/Score/components/ScoreTable/index.tsx
@@ -38,6 +38,24 @@ type StudentsState = {
 
 const courseTasksApi = new CoursesTasksApi();
 
+function trimFiltersProps(filters: ScoreTableFiltersModified): ScoreTableFiltersModified {
+  const trimmedFiltersProps = structuredClone(filters);
+  const filtersKeys = Object.keys(filters) as (keyof ScoreTableFiltersModified)[];
+
+  filtersKeys.forEach(key => {
+    const filterKey = filters[key];
+    if (filterKey && typeof filterKey !== 'boolean') {
+      key = key as Exclude<keyof ScoreTableFiltersModified, 'activeOnly'>;
+      if (Array.isArray(filterKey)) {
+        trimmedFiltersProps[key] = filterKey.map(elem => elem.trim());
+      } else {
+        trimmedFiltersProps[key] = filterKey.trim();
+      }
+    }
+  });
+  return trimmedFiltersProps;
+}
+
 export function ScoreTable(props: Props) {
   const router = useRouter();
   const { width } = useWindowDimensions();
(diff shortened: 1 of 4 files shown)


client/src/modules/Score/components/ScoreTable/index.tsx
```diff
@@ -161,6 +161,24 @@ export function ScoreTable(props: Props) {
     return null;
   }
 
+  function trimFiltersProps(filters: ScoreTableFiltersModified): ScoreTableFiltersModified {
```
I-vasilich-I, line 164: could you move that helper function outside of the component?

client/src/modules/Score/components/ScoreTable/index.tsx
```diff
@@ -161,6 +161,24 @@ export function ScoreTable(props: Props) {
...
+    const trimmedFiltersProps = structuredClone(filters);
+    const filtersKeys = Object.keys(filters) as (keyof ScoreTableFiltersModified)[];
+
+    filtersKeys.forEach(key => {
+      const filterKey = filters[key];
+      if (filterKey && typeof filterKey !== 'boolean') {
+        key = key as Exclude<keyof ScoreTableFiltersModified, 'activeOnly'>;
+        if (typeof filterKey === 'object') {
```
I-vasilich-I, line 172: `Array.isArray()` seems more appropriate to use here

client/src/utils/optionalQueryString.ts
```diff
@@ -1,3 +1,3 @@
-export function optionalQueryString(value: string | undefined) {
+export function optionalQueryString(value: string | string[] | undefined) {
   return value ? String(value) : undefined;
```
I-vasilich-I, line 2: if value could be an array, shouldn't return value be adjusted or it would work fine as it is?

client/src/modules/Score/hooks/types.ts
```diff
@@ -1,8 +1,8 @@
 export type ScoreTableFilters = {
-  githubId?: string;
-  name?: string;
-  'mentor.githubId'?: string;
-  cityName?: string;
+  githubId?: string | string[];
+  name?: string | string[];
+  'mentor.githubId'?: string | string[];
+  cityName?: string | string[];
```
I-vasilich-I, line 5: could you explain how filter could have those properties as arrays?

Zoro444: This is the data we get when we make a filter.

filter-props

client/src/modules/Score/components/ScoreTable/index.tsx
Zoro444: ok I did it.

client/src/modules/Score/components/ScoreTable/index.tsx
Zoro444: ok I did it.

client/src/modules/Score/components/ScoreTable/index.tsx
```diff
@@ -38,6 +38,22 @@ type StudentsState = {
...
+    const filterKey = filters[key];
+    if (filterKey && typeof filterKey !== 'boolean') {
+      key = key as Exclude<keyof ScoreTableFiltersModified, 'activeOnly'>;
+      if (Array.isArray(filterKey)) {
+        trimmedFiltersProps[key] = filterKey.map(elem => elem.trim());
+      }
+    }
+  });
+  return trimmedFiltersProps;
+}
```
AlreadyBored, line 55: Please consider to do it somehow like this
```suggestion
function trimFiltersProps(
  filters: ScoreTableFiltersModified
): ScoreTableFiltersModified {
  const trimmedFiltersProps = structuredClone(filters);

  Object.entries(filters).forEach(([filterKey, filterValue]) => {
    if (filterValue && typeof filterValue !== 'boolean') {
      if (Array.isArray(filterValue)) {
        trimmedFiltersProps[
          filterKey as keyof ScoreTableFiltersModified
        ] = filterValue.map(elem => elem.trim());
      }
    }
  });

  return trimmedFiltersProps;
}
```

valerydluski: Wouldn't it be simpler to fix this here and get rid of this bug in many places at once?
https://github.com/rolling-scopes/rsschool-app/blob/master/client/src/utils/queryParams-utils.ts

client/src/utils/optionalQueryString.ts
```diff
@@ -1,3 +1,7 @@
-export function optionalQueryString(value: string | undefined) {
-  return value ? String(value) : undefined;
+export function optionalQueryString(value: string[] | undefined) {
+  if (Array.isArray(value)) {
+    return String(value[0]);
+  }
+
+  return undefined;
```
AlreadyBored, line 6: You don't need to explicitly return `undefined` as functions return `undefined` by default, please consider to return `null`/don't return at all

client/src/modules/Score/components/ScoreTable/index.tsx
Zoro444: I have already remove this function

client/src/utils/optionalQueryString.ts
```diff
@@ -1,3 +1,7 @@
-export function optionalQueryString(value: string | undefined) {
-  return value ? String(value) : undefined;
+export function optionalQueryString(value: string | string[] | undefined) {
+  if (Array.isArray(value)) {
+    return String(value[0]).trim();
```
aaliakseyenka, line 3: it might be confusing - if you pass an array for serialisation you would expect to serialise it, not taking only first element. same for the fn in file below

Zoro444: we can’t return the array because it’s expected to be string and we get errors
we return the first element of the array since ant design filters gives us the data as an array but with only one element

aaliakseyenka: right, I don't propose to return an array, but rather `serialise` it. ie to map array to string. I understand that antd might return array, but this function doesn't have anything related with antd(it is located in `utils` and makes `query string`. That's why it looks confusing to me.

Zoro444: do I get you correctly that you propose to iterate through the array and make it as a string. Smth like this ```value.reduce((acc, cur) =>  `${acc} ${cur}`,'').trim().```

Zoro444: or do you mean just do a regular join  `array.join('')` ?

aaliakseyenka: I guess the better option would be array.join(',')

client/src/utils/queryParams-utils.ts
```diff
@@ -9,10 +9,10 @@ export const getQueryParams = (
   let params = { ...initialQueryParams };
   for (const [key, value] of Object.entries(queryParams)) {
     if (!isNil(value)) {
-      if (Array.isArray(value) && value[0] !== '') {
-        params = { ...params, [key]: value[0] };
+      if (Array.isArray(value) && value.join(',') !== '') {
+        params = { ...params, [key]: value.join(',').trim() };
```
I-vasilich-I, line 13: it would make more sense to apply `trim` on each element, then `join`, then check it against empty string, then save. Otherwise it won't trim spaces in the middle of joined elements 
`['   it is  ', '   a bad value   '].join(',').trim()` -> `it is  ,   a bad value`

Zoro444: like this?

`  if (Array.isArray(value)) {
        const trimmedElements = value.map(elem => elem.trim()).join(',');
        if (trimmedElements !== '') {
          params = { ...params, [key]: trimmedElements };
        }`

PR #2408, opened 2024-01-18
feat: expose endpoint with endorsement data

apalchys:

---

#### 🤔 This is a ...

- [ ] New feature
- [ ] Bug fix
- [ ] Refactoring
- [ ] Test Case
- [ ] Documentation update
- [ ] Other

#### 🔗 Related issue link

### DIFF
--- a/nestjs/src/profile/profile.controller.ts
+++ b/nestjs/src/profile/profile.controller.ts
@@ -8,11 +8,10 @@ import { ProfileDto } from './dto/profile.dto';
 import { ProfileService } from './profile.service';
 import { PersonalProfileDto } from './dto/personal-profile.dto';
 import { EndorsementService } from './endorsement.service';
-import { EndorsementDto } from './dto/endorsement.dto';
+import { EndorsementDataDto, EndorsementDto } from './dto/endorsement.dto';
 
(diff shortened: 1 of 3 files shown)


valerydluski:
Do we need to run the openapi script?

nestjs/src/profile/profile.controller.ts
```diff
@@ -87,10 +90,17 @@ export class ProfileController {
     return new EndorsementDto(endorsement);
   }
 
+  @Get(':username/endorsement-data')
+  @ApiOperation({ operationId: 'getEndorsementData' })
+  @ApiResponse({ type: EndorsementDto })
```
AlreadyBored, line 95: I believe we need some access guard for this, cause it exposes data to anyone

valerydluski: That's right, as far as I know, that was the plan

AlreadyBored: It could be not a problem if app code was not in open source, but in this case we should make sure that only publicly available data is exposed here, I am not sure that all the stuff here is public

valerydluski (approved):

Issue #2473, opened 2024-05-08
[Course stats] Amount of eligible students is greater than 100%


AlreadyBored:
**Describe the bug**
Course stats display higher amount of eligible students than active students amount, which leads to values higher than 100% in corresponding widget

**To Reproduce**
1. Go to https://app.rs.school/course/stats?course=js-fe-preschool-2024q2
2. See value greater than 100%  
<img width="374" alt="image" src="https://github.com/rolling-scopes/rsschool-app/assets/34455330/5f5a1617-8a7f-4d6a-8b7c-49b4779a2b1c">  

**Expected behavior**
Taking into account that only active users should be eligible for classification, if eligible amount is greater than active students amount => 100% of students are eligible

AlreadyBored:
fixed

PR #2474, opened 2024-05-08
Show correct eligible students amount

AlreadyBored:

---

#### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Refactoring
- [ ] Test Case
- [ ] Documentation update
- [ ] Other

#### 🔗 Related issue link

https://github.com/rolling-scopes/rsschool-app/issues/2473

If amount of eligible students is greater than 100% => it should be 100% (as alternative solution we can show amount of eligible students vs all students, but I think it doesn't make much sense)

### DIFF
--- a/nestjs/src/courses/stats/course-stats.service.ts
+++ b/nestjs/src/courses/stats/course-stats.service.ts
@@ -55,7 +55,7 @@ export class CourseStatsService {
         'students_with_certificate',
       )
       .addSelect(
-        `COUNT(CASE WHEN student.totalScore >= (${maxScore} * course.certificateThreshold / 100) THEN 1 END)`,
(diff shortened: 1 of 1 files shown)


valerydluski:
Are you sure that the problem is not on the backend side?

AlreadyBored:
> Are you sure that the problem is not on the backend side?  

I thought about it, but I am not sure that proper option is to change eligible students calculation on BE, as we may still need overall eligible students amount on FE. But if you are sure that this is the only application of this value, then I can adjust BE calculation

valerydluski:
@AlreadyBored add a filter for active students here.
https://github.com/rolling-scopes/rsschool-app/blob/master/nestjs/src/courses/stats/course-stats.service.ts#L57

valerydluski (approved):

PR #2514, opened 2024-07-19
chore: remove auth guard from get request to courses

Wystov:

---

#### 🤔 This is a ...

- [x] Other

#### 🔗 Related issue link

[rs-site-issue#404](https://github.com/rolling-scopes/site/issues/404)

As a part of integration of courses dates with rs-site, auth guard from get request to api/v2/courses was removed.


apalchys:
@dzmitry-varabei @Wystov Please give us a list of data you need from RS App and we will give you it in a nice format somewhere/somehow.  Direct rs site - rs app integration won't work well especially for internal api.

AlreadyBored:
Not sure it's a good idea to remove auth for some data and expose endpoint for public usage :)

Issue #2625, opened 2025-03-17
Add "Create Interview Pairs" Button to Interview Management Page


dzmitry-varabei:
**Describe the feature**
- The Technical Screening Management page (https://app.rs.school/course/admin/stage-interviews?course=js-fe-2024Q4) currently has a "Create Interview Pairs" button that randomly generates mentor-mentee pairs for interviews.

![Image](https://github.com/user-attachments/assets/004aad8a-984c-49a6-a1a4-e99feff65620)

- The CoreJS Interview Management page (https://app.rs.school/course/admin/interviews?course=js-fe-2024Q4) lacks a similar button.

![Image](https://github.com/user-attachments/assets/09873bc8-dbe0-4fc9-bf80-513fd4ee934f)

- Instead, users are directed to use the "distribute" button on the Tasks Management page (https://app.rs.school/course/admin/tasks?course=js-fe-2024Q4), which is not intuitive.

![Image](https://github.com/user-attachments/assets/c74f4872-53bd-4ef6-9f8e-8e2dd148d8bc)

**Expected Behavior**:
- The CoreJS Interview Management page (https://app.rs.school/course/admin/interviews?course=js-fe-2024Q4) should include a "Create Interview Pairs" button that functions like the one on the Technical Screening Management page, generating random mentor-mentee pairs for interviews.

oibioib:
I will fix

AlreadyBored:
Done

PR #2628, opened 2025-03-17
fix: add Create Interview Pairs button to Interview Management Page

oibioib:

* [X] I followed naming convention rules

---

#### 🤔 This is a ...

- [ ] New feature
- [X] Bug fix
- [ ] Refactoring
- [ ] Test Case
- [ ] Documentation update
- [ ] Other

#### 🔗 Related issue link

- Related Issue #2625

button


client/src/pages/course/admin/interviews.tsx
```diff
@@ -46,20 +48,39 @@ function Page() {
     }
   };
 
+  const createInterviews = async () => {
+    if (selected) {
+      await courseService.createInterviewDistribution(+selected);
```
apalchys, line 53:
```suggestion
      await courseService.createInterviewDistribution(Number(selected));
```

it's better to write conversions in an explicit way

dzmitry-varabei: @valerydluski mentioned today that on courses with a large number of students, creating interview pairs can take a significant amount of time. Without showing a loading screen, users won’t understand that something is in progress and might click the button again. This leads to duplicate pairs being created, which later have to be cleaned up manually in the DB :(

So, we need to:

1) Add a loading screen that stays visible for the entire operation.
2) Leave clear comments for future developers explaining why this is necessary.
3) If possible, refactor the createInterviewDistribution logic to make sure the process can't be triggered multiple times at once.

P.S. Big thanks to @valerydluski Megamind, for his attention and alertness 😎

dzmitry-varabei (changes requested):

AlreadyBored (changes requested):
@oibioib please add changes suggested by @dzmitry-varabei and @apalchys

client/src/pages/course/admin/interviews.tsx
oibioib: Number(selected) fixed.

AlreadyBored:
@oibioib let's do server-side lock as we discussed, current solution is not safe

server/src/migrations/1743946332709-CourseTaskProcessing.ts
```diff
@@ -0,0 +1,13 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class CourseTaskProcessing1743946332709 implements MigrationInterface {
+  name = 'CourseTaskProcessing1743946332709';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`ALTER TABLE "course_task" ADD "isProcessing" boolean NOT NULL DEFAULT false`);
```
AlreadyBored, line 7: IMO `isProcessing` is not the best name for this flag, it's too generic, let's change it to something more semantic

server/src/routes/course/interviews.ts
```diff
@@ -133,11 +125,27 @@ export const createInterviews = (logger: ILogger) => async (ctx: Router.RouterCo
...
+
+    const courseTask = await taskService.getCourseTask(courseTaskId);
+
+    if (courseTask == null) {
+      setResponse(ctx, StatusCodes.BAD_REQUEST, { message: 'not valid course task' });
+      return;
+    }
+
+    if (courseTask.isProcessing) {
+      setResponse(ctx, StatusCodes.PROCESSING, { message: 'course task is already being processed' });
```
AlreadyBored, line 137: `1xx` status codes are informational and shouldn't be used for client error indication, so `Processing (102)` is not a correct status code for this case, please consider using `Conflict (409)` instead

AlreadyBored (changes requested):
@oibioib please take a look at the comments I've left.  
Plus as we already discussed, I remind you that RS App dev team has agreement on kinda "code freeze" in `server` folder, except minor/urgent fixes, this looks like extension of existing logic and it's better to move this functionality to `nestjs` folder just like you did with `getInterviewPairs`.
It can be done as separated PR right after merge of this PR – but only if @apalchys approves such approach, otherwise please consider doing migration to `nestjs` first

server/src/migrations/1743946332709-CourseTaskProcessing.ts
oibioib: Rename to **isCreatingInterviewPairs**.

server/src/routes/course/interviews.ts
oibioib: Fix this

Issue #2642, opened 2025-03-22
Automate rs.school site rebuild after course info update


dzmitry-varabei:
After updating courses.json (https://cdn.rs.school/app/courses.json) (e.g. opening registration or adding a new link like wearecommunity), the https://rs.school/ site is not automatically rebuilt and becomes outdated :( It would be great to automate the rebuilding of https://rs.school/ after any changes in courses.json.

Also, the course registration end date in courses.json should control whether registration remains open on https://rs.school/. After the end date, registration on the site should automatically close.

Maybe it makes sense to rebuild by cron every hour? But if there's a better way, let's use it

PR #2652, opened 2025-03-25
feat: trigger rs.school update when courses changed

apalchys:

---

#### 🤔 This is a ...

- [x] New feature
- [ ] Bug fix
- [ ] Refactoring
- [ ] Test Case
- [ ] Documentation update
- [ ] Other

#### 🔗 Related issue link

#2642 

I moved all logic related to exporting courses and rebuilding rs.school to "ListenerModule" in order to de-couple it from courses service


nestjs/src/listeners/course.listener.ts
```diff
@@ -0,0 +1,118 @@
...
+  /**
+   * Check if courses are changed since last export
+   * @returns true if courses are changed, false otherwise
+   */
+  private async areCoursesChanged() {
+    const response = await this.s3.headObject({ Bucket: this.bucket, Key: this.objectKey });
+    const lastModifiedCourse = await this.repository.findOne({
+      where: this.getCoursesWhereFilter(),
+      order: { updatedDate: 'DESC' },
+    });
```
AlreadyBored, line 53: As I can see these API calls don't depend on each other  
```suggestion
    const [response, lastModifiedCourse] = await Promise.all([
      this.s3.headObject({ Bucket: this.bucket, Key: this.objectKey }),
      this.repository.findOne({
        where: this.getCoursesWhereFilter(),
        order: { updatedDate: 'DESC' },
      }),
    ]);
```

nestjs/src/listeners/course.listener.ts
```diff
@@ -0,0 +1,118 @@
...
+    return (response.LastModified?.getTime() ?? 0) < (lastModifiedCourse?.updatedDate.getTime() ?? 0);
```
AlreadyBored, line 54: In general this is working solution as long as we have one source of updates for s3 object, but IMO it's not very reliable (e.g. object was manually re-uploaded or we'll have 2 ways to update this this object in future – and CDN conent can be mistakenly considered "newer" than courses in DB).  
As an alternative I can suggest following flow:  
1. Get courses from DB and CDN
2. Map courses from DB to `ExportCourseDto`
3. Calculate hash for both sources:  
  a) if hashes are equal – skip CDN update
  b) if hashes differ – update CDN  
This solution is slightly slower, but taking into account that amount of data is small and most likely won't grow significantly in the nearest future + this is background CRON job I think it's acceptable.

apalchys: @AlreadyBored thanks. I can't say I agree with your arguments but i liked the idea :) It makes code simplier. Also I don't do hash - just string comparison. Updated

nestjs/src/listeners/course.listener.ts
apalchys: not actual anymore.

valerydluski (approved):

Issue #2690, opened 2025-04-23
Students can manipulate test results to get 100% or more


dzmitry-varabei:
A vulnerability has been found that allows students to pass auto-tests with 100% score or even more, knowing only a single correct answer.

More details: https://t.me/c/1222826846/10053

woodo01:
Actually, the issue is broader.
It's not even necessary to know a single correct answer — sending any character is enough, and it will be accepted as a valid answer.

For example, when unpacking a value like 0, both `index` and `value` can end up being `undefined`.
https://github.com/rolling-scopes/rsschool-app/blob/master/nestjs/src/courses/task-verifications/self-education.service.ts#L84
So `answers[index]` returns undefined, and the comparison ends up being `undefined === undefined`, which always passes.

Additionally, there's no check on the number of answers — a user can submit even `1000` answers.
As a result, a POST request like `[0, 1, {}]` will yield 3 correct answers.

<img width="379" alt="Image" src="https://github.com/user-attachments/assets/e0a66844-0b76-4935-a586-3968668acfc2" />

This clearly needs to be addressed — validation should be added, along with unit tests.

I wasn’t sure where to report this, since I didn’t know whether it should be a public issue.
I’d be happy to prepare a pull request if that’s okay with you.

PR #2697, opened 2025-05-04
fix: self education validation

apalchys:

**Issue**:
#2690 

**Description**:
Fixing possibility to cheat self-education validation and add test-suite 

**Self-Check**:

- [x] Changes tested locally

### DIFF
--- a/nestjs/src/courses/task-verifications/self-education.service.test.ts
+++ b/nestjs/src/courses/task-verifications/self-education.service.test.ts
@@ -25,7 +25,7 @@ const createMockCourseTask = (
           numberOfQuestions: 2,
           ...attributes.public,
         },
-        answers: attributes.answers,
+        answers: attributes.answers ?? [0, 1],
       },
     } as unknown as Task,
   } as unknown as CourseTask;
@@ -71,28 +71,33 @@ describe('SelfEducationService', () => {
     });
 
     describe('Input Validation', () => {
-      it('should throw if answer count !== numberOfQuestions (less)', () => {
-        studentAnswers = [{ index: 0, value: 0 }];
-
-        mockTask = createMockCourseTask({
-          public: { numberOfQuestions: 2 },
-          answers: [0, 1],
-        });
-
-        expect(() => service.verifySelfEducationAnswers(mockTask, studentAnswers, 0)).toThrow(
-          new BadRequestException('Number of submitted answers (1) does not match the number of questions (2)'),
-        );
-      });
-
-      it('should throw if answer count !== numberOfQuestions (more)', () => {
+      it('should throw if student answers count greater than total answers', () => {
         studentAnswers = [
           { index: 0, value: 0 },
           { index: 1, value: 1 },
           { index: 2, value: 2 },
         ];
-        mockTask = createMockCourseTask({ public: { numberOfQuestions: 2 } });
+
+        mockTask = createMockCourseTask({
+          public: {},
+          answers: [0, 1],
+        });
+
         expect(() => service.verifySelfEducationAnswers(mockTask, studentAnswers, 0)).toThrow(
-          new BadRequestException('Number of submitted answers (3) does not match the number of questions (2)'),
+          new BadRequestException('Incorrect student answers count'),
+        );
+      });
+
+      it('should throw if student answers count is 0', () => {
+        studentAnswers = [];
+
+        mockTask = createMockCourseTask({
+          public: {},
+          answers: [0, 1],
+        });
+
+        expect(() => service.verifySelfEducationAnswers(mockTask, studentAnswers, 0)).toThrow(
+          new BadRequestException('Incorrect student answers count'),
         );
       });
 
@@ -107,7 +112,7 @@ describe('SelfEducationService', () => {
         );
       });
 
-      it('should throw  if input value is undefined', () => {
+      it('should throw if input value is undefined', () => {
         studentAnswers = [
           // @ts-expect-error - test
           { index: 0, value: undefined },
@@ -125,6 +130,39 @@ describe('SelfEducationService', () => {
           { index: 1, value: 1 },
         ];
         mockTask = createMockCourseTask({ public: { numberOfQuestions: 2 } });
+        expect(() => service.verifySelfEducationAnswers(mockTask, studentAnswers, 0)).toThrow(
+          new BadRequestException('Invalid answer value'),
+        );
+      });
+
+      it('should throw if index is out of range', () => {
+        studentAnswers = [
+          { index: 0, value: 0 },
+          { index: 2, value: 1 },
+        ];
+        mockTask = createMockCourseTask({ public: {} });
+        expect(() => service.verifySelfEducationAnswers(mockTask, studentAnswers, 0)).toThrow(
+          new BadRequestException('Invalid answer index'),
+        );
+      });
+
+      it('should throw if index is negative', () => {
+        studentAnswers = [
+          { index: -1, value: 0 },
+          { index: 1, value: 1 },
+        ];
+        mockTask = createMockCourseTask({ public: {} });
+        expect(() => service.verifySelfEducationAnswers(mockTask, studentAnswers, 0)).toThrow(
+          new BadRequestException('Invalid answer index'),
+        );
+      });
+
+      it('should throw if value is negative', () => {
+        studentAnswers = [
+          { index: 0, value: -1 },
+          { index: 1, value: 1 },
+        ];
+        mockTask = createMockCourseTask({ public: {} });
         expect(() => service.verifySelfEducationAnswers(mockTask, studentAnswers, 0)).toThrow(
           new BadRequestException('Invalid answer value'),
         );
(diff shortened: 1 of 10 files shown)


apalchys:
cc @woodo01

nestjs/src/courses/task-verifications/self-education.service.ts
```diff
@@ -63,19 +92,33 @@ export class SelfEducationService {
...
+    if (studentAnswers.flatMap(a => (Array.isArray(a.value) ? a.value : [a.value])).some(a => !Number.isInteger(a))) {
+      throw new BadRequestException('Invalid answer value');
+    }
+
+    const submittedIndices = new Set(studentAnswers.map(a => a.index));
+    if (submittedIndices.size !== numberOfQuestions) {
+      throw new BadRequestException('Submitted answer indices must be unique');
+    }
+
+    for (let i = 0; i < numberOfQuestions; i++) {
```
woodo01, line 111: Sorry, I am a bit confused here.
Indices might be any numbers from 0 to `answers.length`.
It can be 50 `answers`, but `numberOfQuestions` is any 20 of them.

nestjs/src/courses/task-verifications/self-education.service.ts
woodo01, line 1: Probably it makes sense to refactor a bit the file.
const `rightAnswersCount` and `studentCorrectAnswers` have the same logic.
It can be changed for example as:
```
const studentCorrectAnswers: CorrectAnswer[] = studentAnswers.map(({ index, value }) => {
      const sortedAnswers = this.sortAnswers(answers[index]);
      const sortedValues = this.sortAnswers(value);
      return { index, value, isCorrect: sortedAnswers === sortedValues };
    });
const rightAnswersCount =  studentCorrectAnswers.filter((answer: CorrectAnswer) => answer.isCorrect === true).length
```

woodo01: Probably `const studentCorrectAnswers: CorrectAnswer[]` could be renamed as
`const studentAnswers: StudentAnswer[]` because it has incorrect answers as well.

nestjs/src/courses/task-verifications/self-education.service.ts
```diff
@@ -52,6 +46,41 @@ export class SelfEducationService {
...
+      studentId,
+      courseTaskId,
+      score,
+      details,
+      status: 'completed',
+      answers: studentCorrectAnswers,
+    });
+
+    const result = await this.taskVerificationsRepository.findOneByOrFail({ id });
+    await this.writeScoreService.saveScore(studentId, courseTaskId, { score, comment: details });
```
AlreadyBored, line 73: Minor:  
```suggestion
    const [result] = await Promise.all([
      this.taskVerificationsRepository.findOneByOrFail({ id }),
      this.writeScoreService.saveScore(studentId, courseTaskId, { score, comment: details }),
    ]);
```

nestjs/src/courses/task-verifications/self-education.service.ts
AlreadyBored: @woodo01 with current implementation all test questions are mandatory, it's highlighted on FE as well https://github.com/rolling-scopes/rsschool-app/blob/master/client/src/modules/AutoTest/components/SelfEducation/SelfEducation.tsx#L58

AlreadyBored (approved):

nestjs/src/courses/task-verifications/self-education.service.ts
woodo01: Yes, you're right. All FE questions are mandatory. But these questions are taken randomly from the whole list of questions.
https://github.com/rolling-scopes/rsschool-app/blob/master/client/src/modules/AutoTest/components/SelfEducation/SelfEducation.tsx#L22

For example, if there are 50 questions (in a response, see please a screenshot below) and `numberOfQuestions` is 20, 20 randomly taken questions will be shown on FE.
It means indices can be not only in interval [0...19], but they can be [21, 35, 47...].

There is another problem of `numberOfQuestions`. It can be less then the real number of questions.
For example, in the Nodejs Basics test, the `numberOfQuestions` is 30, but the real number of questions is 29.
FE shows 29 questions only and therefore only 29 questions can be sent for verification. But the verification uses `numberOfQuestions` as 30 to calculate accuracy. So the maximum scores can be 29/30. This is why students cannot reach 100% accuracy for the test (they have  291/300 scores maximum). But this can be fixed in backoffice, I guess.
image

AlreadyBored: Hm, probably makes sense. Maybe you can make some suggestions for extra validation/relying on correct indices right in this PR so it would be easier to understand your proposal?

nestjs/src/courses/task-verifications/self-education.service.ts
```diff
@@ -63,19 +92,33 @@ export class SelfEducationService {
...
 
-    const verifications = await this.taskVerificationsRepository.find({
-      where: { studentId, courseTaskId },
-      order: { createdDate: 'DESC' },
-    });
+    // Check if all answer values are integers
+    if (studentAnswers.flatMap(a => (Array.isArray(a.value) ? a.value : [a.value])).some(a => !Number.isInteger(a))) {
+      throw new BadRequestException('Invalid answer value');
+    }
+
```
woodo01, line 105:
```suggestion

    if (studentAnswers.some(a => !Number.isInteger(a.index))) {
      throw new BadRequestException('Invalid answer index');
    }

```
Because everything can be sent by direct post request.
I am not sure btw whether FE sends integers, probably it may send strings like ['1', '2', 3'].
And it will be not valid then.

nestjs/src/courses/task-verifications/self-education.service.ts
```diff
@@ -63,19 +92,33 @@ export class SelfEducationService {
...
+    const submittedIndices = new Set(studentAnswers.map(a => a.index));
+    if (submittedIndices.size !== numberOfQuestions) {
+      throw new BadRequestException('Submitted answer indices must be unique');
+    }
+
+    for (let i = 0; i < numberOfQuestions; i++) {
+      if (!submittedIndices.has(i)) {
+        throw new BadRequestException(`Missing answer for question index ${i}`);
+      }
+    }
```
woodo01, line 115:
```suggestion
    if ([...submittedIndices].some(i => i < 0 || i > answers.length - 1) {
      throw new BadRequestException(`The answer index ${i} is out of range`);
    }
```
Because we don't know exactly which questions from all questions FE showed to a student. Answers here are from task attributes - `courseTask.task.attributes` and they are all answers of all possible questions.

woodo01: The function below could be changed as well by checking for undefined to make double validation.
```
    const rightAnswersCount = studentAnswers
      .map(({ index, value }) => {
        const rightAnswer = this.sortAnswers(answers[index]);
        const userAnswer = this.sortAnswers(value);
        return rightAnswer !== undefined && userAnswer !== undefined && Number(rightAnswer === userAnswer);
      })
      .reduce((sum, value) => sum + value, 0);
```

nestjs/src/courses/task-verifications/self-education.service.ts
apalchys: ok, I removed the initial check but added others. Please review again @AlreadyBored @woodo01

nestjs/src/courses/task-verifications/self-education.service.ts
apalchys: updated the code

nestjs/src/courses/task-verifications/self-education.service.ts
apalchys: since "saveScore" is "complex" and requires additional refactoring, I would leave as is. It's not critical for performance and it's better to execute in the order

nestjs/src/courses/task-verifications/self-education.service.ts
apalchys: added. thanks

woodo01 (approved):

nestjs/src/courses/task-verifications/self-education.service.test.ts
```diff
@@ -0,0 +1,653 @@
...
+        expect(() => service.verifySelfEducationAnswers(mockTask, studentAnswers, 0)).toThrow(
+          new BadRequestException('Submitted answer indices must be unique'),
+        );
+      });
+
+      it('should throw if input value is undefined', () => {
+        studentAnswers = [
+          // @ts-expect-error - test
+          { index: 0, value: undefined },
+          { index: 1, value: 1 },
```
woodo01, line 119: Just a minor comment.
Is it possible to add the following negative cases?
```
// empty object
[
  {},
  { index: 1, value: 1 },
]

// number
[
  123,
  { index: 1, value: 1 },
]

// null
[
  null,
  { index: 1, value: 1 },
]
```
123, null, and {} are invalid for typescript but it is because of type assertion.
But as received body from a client, it can be everything.
Validation is fixed for these cases.

Issue #2691, opened 2025-04-28
Implement Course Prerequisites Validation System


YulikK:
**Describe the feature**
We need to implement a validation system that checks if a student has any RS School certificate before allowing registration for specific courses. Initially, this will be used for the Angular course but should be designed to accommodate similar requirements for other courses in the future.

**Additional context**
Implement a friendly error page when accessing registration link without required certificates
- Display custom message with sloth illustration:
"Dear friend, unfortunately you cannot register for this course as we couldn't find any RS School certificates(or the name of the certificate if we need a specific one) in your profile. Keep an eye on new course starts and join us when you're ready!"
- Add appropriate sloth illustration that matches the message context

AlreadyBored:
Implemented in https://github.com/rolling-scopes/rsschool-app/pull/2698

PR #2698, opened 2025-05-06
feat: add required certificateDisciplines multiple selector to courses

aliakseitokarev:
**Issue**:
#2691 

**Description**:
{EB462B1E-C569-48C0-AE5A-FDCA5C3EA7EF}
{00552633-55FC-4B90-B45B-441220385ED7}

**Self-Check**:

- [x] Changes tested locally


aliakseitokarev:
{88A8BDDB-0F9D-4AC3-A2A4-7902816ADE86}

Only missing certificates are shown

nestjs/src/courses/dto/create-course.dto.ts
```diff
@@ -105,4 +105,9 @@ export class CreateCourseDto {
   @IsOptional()
   @ApiProperty()
   wearecommunityUrl?: string;
+
+  @IsArray()
+  @IsOptional()
+  @ApiProperty()
+  certificateDisciplines?: string[];
```
apalchys, line 112: Why `string`? I thought it should be discipline ID and be `number`

client/src/modules/Registry/hooks/useStudentData/useStudentData.tsx
```diff
@@ -46,6 +48,47 @@ export function useStudentData(githubId: string, courseAlias: string | undefined
     if (courseAlias) {
       const currentCourse = courses.find(course => course.alias === courseAlias);
       const value = registeredForCourses.some(({ id }) => id === currentCourse?.id);
+
+      if (currentCourse) {
+        const disciplineIds = currentCourse.certificateDisciplines;
+        const { data } = (await disciplinesApi.getDisciplinesByIds({ ids: disciplineIds })) as { data: IdName[] };
+
+        if (disciplineIds.includes(0)) {
+          data.push({ id: 0, name: 'any' });
```
apalchys, line 57: Can we treat an empty field as "any" and not add a special menu item?

server/src/models/course.ts
```diff
@@ -120,4 +120,7 @@ export class Course {
 
   @Column({ nullable: true, type: 'varchar' })
   wearecommunityUrl: string | null;
+
+  @Column({ type: 'simple-array', default: '' })
+  certificateDisciplines: string[];
```
apalchys, line 125: Ah, okay, I see. But maybe let's store only the discipline ID, not the array. If we change the discipline name, it should be updated everywhere.

aliakseitokarev: @apalchys, my naming may be unclear. We store only the discipline IDs, which will be updated if we change the discipline name.
It is the same case for multiple selectors as that:
{51823ED5-0CCD-4005-B787-DCE78B7B8B6A}
{FD982361-8966-4B8B-8410-DA62B650E37A}
{BCF0E493-897D-4676-8C8B-FFF7B87E3B1D}

client/src/modules/Registry/hooks/useStudentData/useStudentData.tsx
aliakseitokarev: @apalchys, not we can't. We need an 'empty' case for no certificate required, and 'any' case for any suitable certificate required. Or one of the admins will add 'any' as a discipline.
But thanks, I forgot to handle the case with 'any separately!

aliakseitokarev:
image
image

AlreadyBored (approved):

PR #2686, opened 2025-04-22
fix: remove core interview table bugs after canceling an interview

aliakseitokarev:
#### 🧑‍⚖️ Pull Request Naming Convention

- Title should follow [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/#summary)
- Do not put issue id in title
- Do not put WIP in title. Use [Draft PR](https://github.blog/2019-02-14-introducing-draft-pull-requests/) functionality
- Consider to add `area:*` label(s)

* [x] I followed naming convention rules

---

#### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Performance optimization
- [ ] Refactoring
- [ ] Test Case
- [ ] Documentation update
- [ ] Other

#### 🔗 Related issue link

-  Related issue #2682

#### 💡 Background and solution

{6C5B0437-8D5F-4A4D-AFF4-BECA2A00EC87}

{A36BBFC8-A695-4C7C-9A65-CD4054E90EA5}

#### ☑️ Self Check before Merge

⚠️ Please check all items below before review. ⚠️

- [x] Database migration is added or not needed
- [x] Documentation is updated/provided, or not needed
- [x] Changes are tested locally

PR #2727, opened 2025-06-05
fix: return zero result for interview

oibioib:
**Description**:
Make zero a valid result and status for interview.

**Self-Check**:

- [X] Changes tested locally

### DIFF
--- a/nestjs/src/courses/interviews/interviews.service.ts
+++ b/nestjs/src/courses/interviews/interviews.service.ts
@@ -107,7 +107,11 @@ export class InterviewsService {
       .leftJoin('tc.student', 'student')
       .leftJoin('mentor.user', 'mentorUser')
       .leftJoin('student.user', 'studentUser')
-      .leftJoin(TaskInterviewResult, 'tir', 'tc.studentId = tir.studentId AND tc.courseTaskId = tir.courseTaskId')
+      .leftJoin(
+        TaskInterviewResult,
+        'tir',
+        'tc.studentId = tir.studentId AND tc.courseTaskId = tir.courseTaskId AND tc.mentorId = tir.mentorId',
+      )
       .addSelect([
         'tc.id',
         'tir.score',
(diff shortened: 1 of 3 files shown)


nestjs/src/courses/interviews/interviews.service.ts
```diff
@@ -125,8 +129,8 @@ export class InterviewsService {
 
     return records.map(record => ({
       id: record.tc_id,
-      result: record.tir_score || null,
-      status: record.tir_score ? InterviewStatus.Completed : InterviewStatus.NotCompleted,
+      result: record.tir_score,
+      status: record.tir_score || record.tir_score === 0 ? InterviewStatus.Completed : InterviewStatus.NotCompleted,
```
AlreadyBored, line 133: How it differs from original implementation? `0` is falsy in logical context, so it will lead to `InterviewStatus.NotCompleted` without this `|| record.tir_score === 0` part

server/src/repositories/interview.repository.ts
```diff
@@ -171,53 +167,6 @@ export class InterviewRepository extends AbstractRepository<TaskChecker> {
     });
     return students;
   }
-
-  private async getInterviewPairs(courseTaskId: number): Promise<InterviewPair[]> {
```
AlreadyBored, line 175: Are you sure it's not used somewhere in UI? Please ensure before removal

oibioib: Before migrating getInterviewPairs from the server to NestJS, it was only used in client/src/pages/course/admin/interviews.tsx. Now, it is no longer needed, since we’ve updated this endpoint to use the NestJS API.

nestjs/src/courses/interviews/interviews.service.ts
oibioib: We need to handle the case where an interview with a score of 0 should still be marked as 'Completed'. The condition || record.tir_score === 0 makes that possible. This was added by @aliakseitokarev https://github.com/rolling-scopes/rsschool-app/pull/2686/files

aliakseitokarev: If the score is zero, we need the interview status to be completed. I don't find any mistakes here.

AlreadyBored: nvm, it's correct

AlreadyBored (approved):

aliakseitokarev (approved):

PR #2681, opened 2025-04-19
fix: remove completed and non mentoring courses from invite

aliakseitokarev:
#### 🧑‍⚖️ Pull Request Naming Convention

- Title should follow [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/#summary)
- Do not put issue id in title
- Do not put WIP in title. Use [Draft PR](https://github.blog/2019-02-14-introducing-draft-pull-requests/) functionality
- Consider to add `area:*` label(s)

* [x] I followed naming convention rules

---

#### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Performance optimization
- [ ] Refactoring
- [ ] Test Case
- [ ] Documentation update
- [ ] Other

#### 🔗 Related issue link

- Related Issue #2677

#### 💡 Background and solution

{17D5909E-7704-4663-B70B-F133D07EFCE2}
{80572CD3-E6D8-4579-965C-18740F91B858}
{8F186989-34D5-4B76-B2D8-642DF6FD7CA0}

EPAM ST 2025Q1 without mentoring!
#### ☑️ Self Check before Merge

⚠️ Please check all items below before review. ⚠️

- [x] Database migration is added or not needed
- [x] Documentation is updated/provided, or not needed
- [x] Changes are tested locally

PR #2736, opened 2025-06-19
fix: add name of completed course to mentor registry invite selector

aliakseitokarev:
**Issue**:
After the merge - #2681 courses, which have been completed, the label displays the course ID instead of the course name.
If we want to remove this course from Pre-Selected, it is inconvenient.

image

**Description**:
Now we can read the name of the completed course in the label as well as its modified style and tooltip.
Also, now we don't send a request to the server if the values in the multiselector don't change.
For example, if we delete any course and then add it immediately and click save, or if we delete all courses from the multiselector and add the same ones in a different order.

image
image

**Self-Check**:

- [x] Changes tested locally

### DIFF
--- a/client/src/pages/admin/mentor-registry.tsx
+++ b/client/src/pages/admin/mentor-registry.tsx
@@ -140,11 +140,11 @@ function Page() {
 
   const handleModalSubmit = useCallback(
     async (values: FormData) => {
-      const originalData = modalData?.record?.preselectedCourses?.map(String).sort();
-      const updatedData = values.preselectedCourses?.map(String).sort();
+      const originalSortedData = modalData?.record?.preselectedCourses.map(courseId => String(courseId)).sort();
+      const updatedPreselectedCourses = values.preselectedCourses.map(courseId => String(courseId));
+      const updatedSortedData = updatedPreselectedCourses.sort();
 
-      const isSame =
-        originalData?.length === updatedData?.length && originalData.every((value, i) => value === updatedData[i]);
+      const isSame = JSON.stringify(originalSortedData) === JSON.stringify(updatedSortedData);
 
       if (isSame) {
         setModalData(null);
@@ -155,7 +155,7 @@ function Page() {
         setModalLoading(true);
         if (modalData?.record?.githubId) {
           await mentorRegistryService.updateMentor(modalData.record.githubId, {
-            preselectedCourses: values.preselectedCourses.map(v => String(v)),
+            preselectedCourses: updatedPreselectedCourses,
           });
         }
         setModalData(null);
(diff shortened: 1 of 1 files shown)


AlreadyBored:
@aliakseitokarev please elaborate: 
1) What issue we fix
2) What is the essence of fix and how it is related with your changes. 

Now I don't understand it from code

aliakseitokarev:
@AlreadyBored, I've added details to the issue and pr description sections

client/src/pages/admin/mentor-registry.tsx
```diff
@@ -140,6 +140,17 @@ function Page() {
 
   const handleModalSubmit = useCallback(
     async (values: FormData) => {
+      const originalData = modalData?.record?.preselectedCourses?.map(String).sort();
+      const updatedData = values.preselectedCourses?.map(String).sort();
```
AlreadyBored: 1) Let's avoid implicit conversions  
```suggestion
      const originalData = modalData?.record?.preselectedCourses?.map(courseId => String(courseId)).sort();
      const updatedData = values.preselectedCourses?.map(courseId => String(courseId)).sort();
```

client/src/pages/admin/mentor-registry.tsx
```diff
@@ -140,6 +140,17 @@ function Page() {
...
+
+      const isSame =
+        originalData?.length === updatedData?.length && originalData.every((value, i) => value === updatedData[i]);
```
AlreadyBored: 2) There's a simplier way to compare these values  
```suggestion
      const isSame = JSON.stringify(originalData) === JSON.stringify(updatedData);
```

client/src/pages/admin/mentor-registry.tsx
AlreadyBored: And let's reuse updatedData here instead of doing this mapping twice
image

client/src/pages/admin/mentor-registry.tsx
```diff
@@ -164,6 +175,13 @@ function Page() {
     if (!data) {
       return null;
     }
+
+    const allShownCourses = courses.filter(
+      course =>
+        (course.completed && data.preselectedCourses.includes(course.id)) ||
+        (!course.completed && course.personalMentoring),
+    );
```
AlreadyBored: Let's put conditions into semantically named variables
```suggestion
    const allShownCourses = courses.filter(course => {
      const isCompletedAndPreselected = course.completed && data.preselectedCourses.includes(course.id);
      const isActiveWithPersonalMentoring = !course.completed && course.personalMentoring;

      return isCompletedAndPreselected || isActiveWithPersonalMentoring;
    });
```

AlreadyBored (approved):

PR #2748, opened 2025-07-06
fix: change telegram link on the mentor's dashboard page

aliakseitokarev:
For megre after: #2750

**Issue**:
#2737 

**Description**:
Now you can access the right course FAQ telegram chat link on the mentor's dashboard page.

image

**Self-Check**:

- [x] Changes tested locally

### DIFF
--- a/client/src/modules/Mentor/components/Instructions/Instructions.tsx
+++ b/client/src/modules/Mentor/components/Instructions/Instructions.tsx
@@ -1,12 +1,42 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import { Badge, Card, Col, Row, Space, Typography } from 'antd';
 import { INSTRUCTIONS_TEXT, renderDescription, renderSocialLinks } from '.';
+import { DiscordServersApi } from 'api';
+
+interface InstructionsProps {
+  discordServerId: number;
+}
 
 const { Meta, Grid } = Card;
 const { Text } = Typography;
 
-function Instructions() {
-  const { title, description, steps } = INSTRUCTIONS_TEXT;
+const discordServer = new DiscordServersApi();
+
+function Instructions({ discordServerId }: InstructionsProps) {
+  const { title, description } = INSTRUCTIONS_TEXT;
+  const [steps, setSteps] = useState(INSTRUCTIONS_TEXT.steps);
+
+  useEffect(() => {
+    if (!discordServerId) return;
+
+    async function fetchTelegramLink() {
+      const discordServerData = await discordServer.getDiscordServerById(discordServerId);
+
+      const updatedSteps = steps.map(step => {
+        if (!step.links) return step;
+
+        const updatedLinks = step.links.map(link =>
+          link.title === 'telegram' ? { ...link, url: discordServerData.data } : link,
+        );
+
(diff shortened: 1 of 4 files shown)


client/src/modules/Mentor/components/Instructions/Instructions.tsx
```diff
@@ -1,12 +1,42 @@
...
+const discordServer = new DiscordServersApi();
+
+function Instructions({ discordServerId }: InstructionsProps) {
+  const { title, description } = INSTRUCTIONS_TEXT;
+  const [steps, setSteps] = useState(INSTRUCTIONS_TEXT.steps);
+
+  useEffect(() => {
+    if (!discordServerId) return;
+
+    async function fetchTelegramLink() {
```
AlreadyBored: This function is not reused and called only once within the same scope. Please put it's logic inside `useEffect` callback without wrapping it in function

client/src/modules/Mentor/components/Instructions/Instructions.tsx
```diff
@@ -1,12 +1,42 @@
...
+      const discordServerData = await discordServer.getDiscordServerById(discordServerId);
+
+      const updatedSteps = steps.map(step => {
+        if (!step.links) return step;
+
+        const updatedLinks = step.links.map(link =>
+          link.title === 'telegram' ? { ...link, url: discordServerData.data } : link,
```
AlreadyBored: I don't fully understand logic here: if link is **telegram** link, then we add **discord** data. Why?

client/src/modules/Mentor/components/Instructions/renderers.tsx
```diff
@@ -21,22 +21,24 @@ const getSocialLinkIcon = (title: string) => {
...
 
 export const renderDescription = (text: string) => {
   return (
     <Text>
-      <div
-        dangerouslySetInnerHTML={{
-          __html: text,
-        }}
-      ></div>
+      <div dangerouslySetInnerHTML={{ __html: text }}></div>
```
AlreadyBored:
```suggestion
      <div dangerouslySetInnerHTML={{ __html: text }} />
```

client/src/modules/Mentor/components/Instructions/Instructions.tsx
aliakseitokarev: @AlreadyBored, I can not use await inside useEffect directly. Or do you mean using useRequest instead? For now, I've replaced it with an arrow function for better readability.
Here is an example of using useEffect in the rs app with arrow function (client\src\components\Forms\Heroes\index.tsx) :

image

UPD: Thanks for the tip. I used useAsyncEffect.

client/src/modules/Mentor/components/Instructions/Instructions.tsx
aliakseitokarev: My naming is still far from normal, but I'm working on it. I'll fix it to make the code self-documenting!

AlreadyBored:
@aliakseitokarev please check build issues

aliakseitokarev:
> @aliakseitokarev please check build issues

image
The build will be good.

AlreadyBored (approved):

PR #2756, opened 2025-07-16
fix: delete session cache after mentor registration

aliakseitokarev:
**Issue**:
There is no access to the FAQ mentor chat invite link after registering as a new mentor.

image

[Screencast from 15-07-25 18_44_54.webm](https://github.com/user-attachments/assets/4f56560c-9919-49e4-a74e-6831feef8f69)

**Description**:
Add a new endpoint to clear the auth user session cache with a frontend request.
Clear user session cache after mentor's registration

[clear user session cache.webm](https://github.com/user-attachments/assets/671dbe02-ab32-4032-be68-068080cac796)

**Self-Check**:

- [x] Changes tested locally


AlreadyBored:
From my point of view looks fine, but maybe there's a better way to deal with it and I am not aware about it.

@apalchys what do you think?

apalchys (approved):

PR #2817, opened 2025-09-14
fix(paginate): skip could not be negative

apalchys:

**Issue**:
Fixing "OFFSET must not be negative" error

**Description**:
Making skip max value as 0

**Self-Check**:

- [x] Changes tested locally

### DIFF
--- a/nestjs/src/auth/auth.service.ts
+++ b/nestjs/src/auth/auth.service.ts
@@ -78,7 +78,9 @@ export class AuthService {
     }
 
     if (result == null) {
-      const [email] = profile.emails?.filter((email: any) => !!email.primary) ?? [];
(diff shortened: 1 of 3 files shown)


nestjs/src/auth/auth.service.ts
```diff
@@ -78,7 +78,9 @@ export class AuthService {
     }
 
     if (result == null) {
-      const [email] = profile.emails?.filter((email: any) => !!email.primary) ?? [];
+      // Github has "primary" field
+      const emails: (NonNullable<Profile['emails']>[number] & { primary?: boolean })[] = profile.emails ?? [];
```
apalchys, line 82: fixing "any"

Issue #2840, opened 2025-10-13
A confirmation window should be added before closing the task editor.


YulikK:
**Describe the feature**
When editing a task, for example, when manually filling out a cross-check form, if you accidentally click outside the window, the window closes without saving the changes.

**Additional context**

<img width="831" height="984" alt="Image" src="https://github.com/user-attachments/assets/f626562d-5a71-4189-9de6-519388a4c291" />

PR #2843, opened 2025-10-14
A confirmation window should be added before closing the task editor

pestler:
**Issue**:
#2840 
Screeshots
image

**Description**:
   The ModalForm component now includes an optional `resetOnCancel` prop to control whether the form is reset upon cancellation.
   Previously, the form was always reset, which prevented components like `TaskModal` from implementing asynchronous cancel logic (e.g., a  confirmation dialog for unsaved changes).   
   Added `resetOnCancel?: boolean` to `ModalForm`, defaulting to `true` to ensure backward compatibility.
   Updated `TaskModal` to use `ModalForm` with `resetOnCancel={false}`, allowing the "unsaved changes" confirmation to function correctly.

**Self-Check**:

- [x] Changes tested locally

### DIFF
--- a/client/src/components/Footer/Feedback.tsx
+++ b/client/src/components/Footer/Feedback.tsx
@@ -1,24 +1,22 @@
 import React from 'react';
-import LikeTwoTone from '@ant-design/icons/LikeTwoTone';
-import HeartTwoTone from '@ant-design/icons/HeartTwoTone';
-import TrophyTwoTone from '@ant-design/icons/TrophyTwoTone';
 import { Menu } from './Menu';
+import { HeartOutlined, LikeOutlined, TrophyOutlined } from '@ant-design/icons';
 
 const publicRoutes = [
   {
-    icon: <LikeTwoTone />,
+    icon: <LikeOutlined />,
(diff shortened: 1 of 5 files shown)


pestler:
The old ModalForm.tsx component isn't suitable for UI improvements. Its cancel handler tightly controls change detection and reacts to the close icon click. If we rewrite it, it will break everything that depends on it — and there are 11 components using it.

So we’re left with two options:

Build without it and use plain Ant Design forms directly.

Create a copy of ModalForm.tsx, rename it, and use the new version instead.

What’s the best way to approach this?

pestler:
done

apalchys:
@pestler @AlreadyBored Task modal stoped working after this change. Modal don't open. 

After a complaint from @YulikK, I reverted this PR and implemented a much simpler change that is very similar in behavior and fixes the original bug.

Please see: https://github.com/rolling-scopes/rsschool-app/commit/86e4a7de542013e95dd325cf8aa4e837ce87c704

We pushed directly to the master branch because we don't have time for reviews during weekends due to a deadline on @YulikK side

Issue #2846, opened 2025-10-15
Need statistics for reason to leave the course


YulikK:
**Describe the feature**
We need to add a window for those who leave the course so that we can understand the statistics and motivation of people who leave the course. I think it would be good to have a short list of checkboxes with reasons that caused a person to drop out of the course. We can also add a text field if a student wants to give a more detailed answer. I also think it is important to duplicate these reasons in Russian and English so as not to create additional difficulties, so that students do not skip the form because of the need to understand and translate. We also need the ability to view analytics on students and courses.

pestler:
<img width="467" height="601" alt="Image" src="https://github.com/user-attachments/assets/135175e9-cb67-4e2b-bb32-115cd6d1357a" />

PR #2854, opened 2025-10-22
feat: add reports page

pestler:
**Issue**:
#2846 

**Description**:
Describe the feature
We need to add a window for those who leave the course so that we can understand the statistics and motivation of people who leave the course. I think it would be good to have a short list of checkboxes with reasons that caused a person to drop out of the course. We can also add a text field if a student wants to give a more detailed answer. I also think it is important to duplicate these reasons in Russian and English so as not to create additional difficulties, so that students do not skip the form because of the need to understand and translate. We also need the ability to view analytics on students and courses.

image
image
image

**Self-Check**:

- [x] Changes tested locally


apalchys:
Minor: https://github.com/rolling-scopes/rsschool-app/blob/master/CONTRIBUTING.md#pull-requests

please use https://github.com/rolling-scopes/rsschool-app/blob/master/CONTRIBUTING.md#git-commit-messages for PR names

client/src/components/Profile/StudentStatsCard.tsx
```diff
@@ -70,10 +126,28 @@ class StudentStatsCard extends React.Component<Props, State> {
...
+        body: JSON.stringify(surveyData),
+      });
+
+      if (!response.ok) {
+        throw new Error('Failed to submit survey');
+      }
+      window.location.reload();
+    } catch (error) {
+      console.error('Failed to submit survey or leave course:', error);
+      this.setState({ isLoading: false });
```
AlreadyBored, line 136: It would be better to put this into `finally` block IMO

client/src/modules/CourseManagement/components/ExpelledStudentsStats.tsx
```diff
@@ -0,0 +1,237 @@
...
+const fetchExpelledStats = async (): Promise<DetailedExpelledStat[]> => {
+  const response = await fetch('/api/course/stats/expelled');
+  if (!response.ok) {
+    throw new Error('Failed to fetch stats');
+  }
+  return response.json() as Promise<DetailedExpelledStat[]>;
+};
+
+const ExpelledStudentsStats: React.FC = () => {
+  const { data, error, loading, refetch } = useRequest(fetchExpelledStats) as unknown as UseRequestResultWithRefetch;
```
AlreadyBored: Why not to use existing `refresh` API call? You shouldn't override `useRequest` result type
```suggestion
  const { data, error, loading, refresh } = useRequest(fetchExpelledStats);
```

client/src/modules/CourseManagement/components/ExpelledStudentsStats.tsx
```diff
@@ -0,0 +1,237 @@
...
+  refetch: () => void;
+}
+
+const fetchExpelledStats = async (): Promise<DetailedExpelledStat[]> => {
+  const response = await fetch('/api/course/stats/expelled');
+  if (!response.ok) {
+    throw new Error('Failed to fetch stats');
+  }
+  return response.json() as Promise<DetailedExpelledStat[]>;
+};
```
AlreadyBored: Let's create hook `useExpelledStats`, that will contain this logic, call `useRequest` inside and return what you need from `useRequest`. You can use it inside of `ExpelledStudentsStats` component

client/src/modules/CourseManagement/components/ExpelledStudentsStats.tsx
```diff
@@ -0,0 +1,237 @@
...
+          {githubId}
+        </a>
+      ),
+    },
+    {
+      title: 'Reasons for Leaving',
+      dataIndex: 'reasonForLeaving',
+      key: 'reasons',
+      render: (reasons: string[]) => {
+        if (!reasons) {
```
AlreadyBored: Is it really possible not to have `reasons` taking into account that it's defined as `string[]`?

client/src/modules/CourseManagement/components/ExpelledStudentsStats.tsx
```diff
@@ -0,0 +1,237 @@
...
+      dataIndex: 'submittedAt',
+      key: 'date',
+      render: dateUtcRenderer,
+    },
+    {
+      title: 'Action',
+      key: 'action',
+      render: (_text, record) => (
+        <Button danger onClick={() => handleDelete(record.id)} loading={isDeleting}>
+          Del
```
AlreadyBored:
```suggestion
          Delete
```

client/src/modules/CourseManagement/components/ExpelledStudentsStats.tsx
```diff
@@ -0,0 +1,237 @@
...
+        </Button>
+      ),
+    },
+  ];
+
+  if (error) {
+    return <p>Failed to load statistics.</p>;
```
AlreadyBored:
```suggestion
    return <Typography.Paragraph>Failed to load statistics.</Typography.Paragraph>;
```
or
```suggestion
    return <Typography.Text>Failed to load statistics.</Typography.Text>;
```
depending on your needs

client/src/modules/CourseManagement/components/ExpelledStudentsStats.tsx
```diff
@@ -0,0 +1,237 @@
...
+          } else if (col.key === 'githubId') {
+            value = row.user ? row.user.githubId : '';
+          }
+          if (value.includes(',') || value.includes('"')) {
+            return `"${value.replace(/"/g, '""')}"`;
+          }
+          return value;
+        })
+        .join(',');
+    });
```
AlreadyBored: This contains a lot of inline logic. Let's make it more readable by splitting into smaller functions with semantic naming, something like  
```typescript
  const escapeCSVValue = (value: string): string => {
    if (value.includes(',') || value.includes('"')) {
      return `"${value.replace(/"/g, '""')}"`;
    }
    return value;
  };

  const getValueFromDataIndex = (row: DetailedExpelledStat, dataIndex: unknown): string => {
    if (Array.isArray(dataIndex)) {
      let current: unknown = row;
      
      for (const key of dataIndex) {
        const keyAsString = String(key);
        current = current ? (current as Record<string, unknown>)[keyAsString] : undefined;
      }
      
      return current !== undefined && current !== null ? String(current) : '';
    }
    
    if (typeof dataIndex === 'string' || typeof dataIndex === 'number') {
      const value = row[dataIndex as keyof DetailedExpelledStat];
      return value !== undefined && value !== null ? String(value) : '';
    }
    
    return '';
  };

  const getSpecialColumnValue = (row: DetailedExpelledStat, columnKey: string): string => {
    switch (columnKey) {
      case 'reasons':
        return row.reasonForLeaving ? row.reasonForLeaving.map(r => r.replace(/_/g, ' ')).join('; ') : '';
      case 'date':
        return row.submittedAt ? new Date(row.submittedAt).toLocaleString() : '';
      case 'courseName':
        return row.course ? row.course.alias : '';
      case 'githubId':
        return row.user ? row.user.githubId : '';
      default:
        return '';
    }
  };

  const formatRowToCsv = (row: DetailedExpelledStat, exportableColumns: ColumnType<DetailedExpelledStat>[]): string => {
    return exportableColumns
      .map(col => {
        let value = '';
        
        if (col.dataIndex) {
          value = getValueFromDataIndex(row, col.dataIndex);
        } else if (col.key) {
          value = getSpecialColumnValue(row, String(col.key));
        }
        
        return escapeCSVValue(value);
      })
      .join(',');
  };
```
And, probably, it should be put outside of the component (or at least wrapped in `useCallback`

client/src/modules/CourseManagement/components/ExpelledStudentsStats.tsx
```diff
@@ -0,0 +1,237 @@
...
+    const url = URL.createObjectURL(blob);
+    setCsvUrl(url);
+
+    setTimeout(() => {
+      if (downloadRef.current) {
+        downloadRef.current.click();
+        URL.revokeObjectURL(url);
+        setCsvUrl(null);
+      }
+    }, 0);
```
AlreadyBored, line 166: I don't get why do we need this trick with zero `setTimeout` and `downloadRef` (and why do we need `downloadRef` as well here, it looks like some kind of not very reliable hack. Could you please clarify it and consider alternative approaches?

client/src/modules/CourseManagement/components/ExpelledStudentsStats.tsx
```diff
@@ -0,0 +1,237 @@
...
+  };
+
+  return (
+    <div style={{ marginTop: 24 }}>
+      <div style={{ display: 'flex', justifyContent: 'space-between', alignItems: 'center', marginBottom: 16 }}>
```
AlreadyBored: Please try to use Antd components like `Space` instead of raw HTML + css

server/src/migrations/index.ts
```diff
@@ -61,6 +61,7 @@ import { Course1736458672717 } from './1736458672717-Course';
 import { CoursePersonalMentoringDates1738250779923 } from './1738250779923-CoursePersonalMentoringDates';
 import { CourseTaskInterviewCreatingPairs1747380525126 } from './1747380525126-CourseTaskInterviewCreatingPairs';
 import { Course1746467689328 } from './1746467689328-Course';
+import { AddCourseLeaveSurveyResponse1760699701354 } from './1760699701354-AddCourseLeaveSurveyResponse';
 
 export const migrations = [
```
AlreadyBored, line 65: Why so many changes in order of existing migrations? You only add a new one, it's expected that it should be simply added on top of existing collection

server/src/routes/course/stats.ts
```diff
@@ -0,0 +1,67 @@
...
+import { ILogger } from '../../logger';
+import { CourseLeaveSurveyResponse } from '../../models';
+import Router from '@koa/router';
+
+function getExpelledStats(logger: ILogger) {
+  return async (ctx: any) => {
+    logger.info('Fetching detailed expelled students statistics');
+    const surveyRepository = getRepository(CourseLeaveSurveyResponse);
+
+    try {
```
AlreadyBored: I don't really think that we need this `try/catch` stuff, why do we expect it to fail?

server/src/routes/course/stats.ts
```diff
@@ -0,0 +1,67 @@
...
+  };
+}
+
+function deleteExpelledStat(logger: ILogger) {
+  return async (ctx: any) => {
+    const { id } = ctx.params;
+    logger.info(`Deleting expelled student stat with ID: ${id}`);
+    const surveyRepository = getRepository(CourseLeaveSurveyResponse);
+
+    try {
```
AlreadyBored: I don't really think that we need this `try/catch` stuff, why do we expect it to fail?

server/src/routes/course/survey.ts
```diff
@@ -0,0 +1,39 @@
...
+      const newSurveyResponse = surveyRepository.create({
+        userId,
+        courseId,
+        reasonForLeaving,
+        otherComments,
+      });
+
+      await surveyRepository.save(newSurveyResponse);
+
+      const comment = `${SELF_EXPELLED_MARK}. Reason: ${reasonForLeaving?.join(', ') || 'N/A'}. Comment: ${otherComments || 'N/A'}`;
```
AlreadyBored: Reasons can be multiple, so
```suggestion
      const comment = `${SELF_EXPELLED_MARK}. Reasons: ${reasonForLeaving?.join(', ') || 'N/A'}. Comment: ${otherComments || 'N/A'}`;
```

server/src/models/course-leave-survey-response.ts
```diff
@@ -0,0 +1,32 @@
...
+
+  @ManyToOne(() => Course)
+  @JoinColumn({ name: 'courseId' })
+  course: Course;
+
+  @Column({ type: 'text', array: true, nullable: true })
+  reasonForLeaving: string[];
+
+  @Column({ type: 'text', nullable: true })
+  otherComments: string;
```
AlreadyBored: If this is a single string value, then it should be
```suggestion
  otherComment: string;
```

AlreadyBored (changes requested):
@pestler thank you for your work I left inline comments and besides we need to migrate this new functionality to `nestjs` folder. It has many fancy stuff out of the box, you won't need all these tricks with manually setting certain response code on error, using `fetch` and so on. `server` code is on code freeze, except critical fixes, it should have only database schema changes and shared models

client/src/components/Profile/StudentStatsCard.tsx
pestler: done

client/src/modules/CourseManagement/components/ExpelledStudentsStats.tsx
pestler: done

client/src/modules/CourseManagement/components/ExpelledStudentsStats.tsx
pestler: done

server/src/routes/course/survey.ts
pestler: done

client/src/modules/CourseManagement/components/ExpelledStudentsStats.tsx
pestler: done

server/src/migrations/index.ts
pestler: done

server/src/models/course-leave-survey-response.ts
pestler: done

server/src/routes/course/stats.ts
pestler: done

AlreadyBored:
@pestler thank you for fixes, please resolve merge conflicts and we're good to go with it

pestler:
Done

PR #2870, opened 2025-11-11
fix: restore access to confirm mentor registration

aliakseitokarev:

**Issue**:
Mentor can't accept the invitation.

**Description**:
Add the mentorConfirmRoute to ActiveCourseProvider, incorporating the necessary checks.

**Self-Check**:

- [x] Changes tested locally


client/src/modules/Course/contexts/ActiveCourseContext.tsx
```diff
@@ -58,6 +62,28 @@ export const ActiveCourseProvider = ({ children, publicRoutes }: Props) => {
...
 
+  const { loading: mentorLoading } = useRequest(
+    async () => {
+      const mentor = await new MentorRegistryService().getMentor();
+      if (!mentor) {
+        return false;
+      }
+      const courseToConfirm = await new CoursesApi().getCourseByAlias(alias as string);
+      const isInvited = mentor?.preselectedCourses?.includes(courseToConfirm?.data.id) ?? false;
+      return isInvited;
```
emp74ark, line 73: How about:
```ts
return !!mentor?.preselectedCourses?.includes(courseToConfirm?.data.id);
```

client/src/pages/_app.tsx
```diff
@@ -23,7 +23,8 @@ class RsSchoolApp extends App {
         <ThemeProvider>
           <MessageProvider>
             <ActiveCourseProvider
-              publicRoutes={['/login', '/registry/mentor', '/registry/student', '/course/mentor/confirm']}
+              publicRoutes={['/login', '/registry/mentor', '/registry/student']}
+              mentorConfirmRoute="/course/mentor/confirm"
```
emp74ark, line 27: This route was excluded from the publicRoutes, but some children render depends on publicRoutes:
```ts
   if (isPublicRoute && router.isReady) {
     return <>{children}</>;
   }
```
Is it expected?

nestjs/src/courses/courses.controller.ts
```diff
@@ -48,6 +48,16 @@ export class CoursesController {
     return new CourseDto(created);
   }
 
+  @Get('/:courseAlias')
+  @ApiOperation({ operationId: 'getCourseByAlias' })
+  @ApiForbiddenResponse()
+  @ApiOkResponse({ type: CourseDto })
+  @UseGuards(DefaultGuard)
+  public async getByAlias(@Req() _: CurrentRequest, @Param('courseAlias') courseAlias: string) {
```
emp74ark, line 56: Does the request object used here?

AlreadyBored (approved):

AlreadyBored:
@aliakseitokarev  please resolve code review concers and merge conflicts and we're good to go :)

AlreadyBored:
Closing as **superseded by #2871**.

While reviewing/rebasing this PR against current `master`, I found that the issue it targets — *"Mentor can't accept the invitation"* — is **already fixed on `master`** by the fast-fix that landed in #2871 (commit `818f05ed`), which simply adds `/course/mentor/confirm` to `publicRoutes`. That route is still public on current `master`, and the confirm page (`client/src/pages/course/mentor/confirm.tsx`) is self-contained: it does its own session check and `preselectedCourses` validation, so making the route public is not insecure.

This PR takes a different approach (move the route out of `publicRoutes` and gate it through `ActiveCourseProvider` via an `isMentorInvited` check). After rebasing onto `master` I verified it and it has a couple of problems in its current form:

1. **Regression for the primary case.** The provider computes `isMentorInvited` but never uses it to actually render the confirm page — it only suppresses the "No Access" alert. For a freshly-invited mentor with no existing course enrollments, `resolveCourse` returns `[null, []]`, so the `data[0] === null` branch renders `WelcomeCard` **instead of the confirm page**. That's exactly the user who is trying to accept their first invitation.
2. **Route collision.** `@Get('/:courseAlias')` was registered before `@Get('/:courseId')`, so it shadowed the existing `getCourse` endpoint (`GET /courses/:id` → treated as an alias lookup → fails). `getCourse` is used by `CourseModal`.
3. The existing `ActiveCourseProvider` unit tests need updating for the new mentor request.

Given the bug is already resolved on `master` and the confirm page guards itself, I don't think the extra `ActiveCourseProvider` machinery + the new `getCourseByAlias` endpoint are needed. Closing this out; happy to reopen if we decide we want the stricter (non-public) gating instead of the fast-fix — in which case points (1)–(3) above need to be addressed first.

PR #2874, opened 2025-11-14
fix: resolve CourseStatistics types issues

aliakseitokarev:

**Issue**:
The data types in StatCards and useCourseStats do not match the types in api.ts after running the npm run openapi script.

**Description**:
Add necessary descriptions for OpenAPI and fix type mismatches.

**Self-Check**:

- [x] Changes tested locally

### DIFF
--- a/nestjs/src/courses/stats/course-stats.controller.ts
+++ b/nestjs/src/courses/stats/course-stats.controller.ts
@@ -52,8 +52,8 @@ export class CourseStatsController {
   @ApiOkResponse({ type: CourseAggregateStatsDto })
   public async getCoursesStats(
     @Req() req: CurrentRequest,
-    @Query('ids', new ParseArrayPipe({ items: Number, optional: true })) ids: number[],
-    @Query('year', new ParseIntPipe({ optional: true })) year: number,
+    @Query('ids', new ParseArrayPipe({ items: Number })) ids: number[],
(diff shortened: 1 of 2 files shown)


nestjs/src/courses/stats/course-stats.controller.ts
```diff
@@ -37,6 +37,18 @@ export class CourseStatsController {
   @CacheTTL(ONE_HOUR_CACHE_TTL)
   @UseInterceptors(CacheInterceptor)
   @ApiOperation({ operationId: 'getCoursesStats' })
+  @ApiQuery({
+    name: 'ids',
+    required: true,
```
sergeyshalyapin, line 42: This field does not match the query requirements.
We have
`@Query('ids', new ParseArrayPipe({ items: Number, optional: true })) ids: number[],`
which means if do not have the parameters we generate `undefined` response.
That means we have possibility to send and accept undefined/null vaules within nest contorller. There are two options:
1. To make the parameters optional, like `ids?: [number], year?: number` and remove `required: true,`.
2. Remove optional statatus for the query and remove `| undefined` for the fetch request answer. But, error handling should be provided too.

I hope this requirement will not confuse you, please, let me know if my review has mismatches.

aliakseitokarev: @sergeyshalyapin, thanks for the review. I chose the second option.

PR #2877, opened 2025-12-01
feat: duplicate student from another course

pestler:
Duplicate student from another course

course RS 2020 Q1
image

another course - TEST COURSE

image

**Description**:
Course ID check added

**Self-Check**:

- [x] Changes tested locally

### DIFF
--- a/client/src/modules/Mentor/pages/Students/index.tsx
+++ b/client/src/modules/Mentor/pages/Students/index.tsx
@@ -26,7 +26,6 @@ export function Students() {
 
   return (
     <PageLayoutSimple title="Your students" loading={loading}>
-
       {students?.length ? (
         students.map(student => {
           const [feedback] = student.feedbacks;


client/src/modules/Mentor/hooks/useMentorStudents.tsx
```diff
@@ -1,26 +1,28 @@
...
 
   const fetchStudents = useCallback(async () => {
-    if (mentorId) {
+    if (mentorId && courseId) {
       setLoading(true);
       try {
-        const { data = [] } = await service.getMentorStudents(mentorId);
-        setStudents(data);
+        const { data = [] } = await service.getMentorStudents(mentorId, courseId);
+        const uniqueStudents = uniqBy(data, 'githubId');
```
AlreadyBored, line 17: @pestler do we really need this deduplication on FE taking into account that we search for mentees inside of specific course on BE?

nestjs/src/courses/mentors/mentors.controller.ts
```diff
@@ -53,17 +53,21 @@ export class MentorsController {
     return new MentorOptionsDto(mentor);
   }
 
-  @Get('/:mentorId/students')
+  @Get('/:mentorId/course/:courseId/students')
```
apalchys, line 56: i think this is wrong. Mentor <-> Course is one to one relationship.

You can get course id from getting mentor from DB.

nestjs/src/courses/mentors/mentors.service.ts
```diff
@@ -84,12 +84,13 @@ export class MentorsService {
...
   }
 
-  public getStudents(mentorId: number, userId: number) {
+  public getStudents(mentorId: number, userId: number, courseId: number) {
     return this.studentRepository
       .createQueryBuilder('student')
       .leftJoinAndSelect('student.user', 'user')
       .leftJoinAndSelect('student.feedbacks', 'feedback', 'feedback.auhtorId = :userId', { userId })
       .where('student.mentorId = :mentorId', { mentorId })
+      .andWhere('student.courseId = :courseId', { courseId })
```
apalchys, line 93: I think we are fixing bad data. If this bug is reproducible only locally, we need to fix the data. Technically, it should not be possible to get students from a different course in this query.

pestler: This problem is observed only in the test database.
I am closing the PR (Pull Request).

PR #2889, opened 2025-12-09
fix: render all student's interview cards with result

aliakseitokarev:
**Issue**:
The student can only see the result of the last interview. If the manager creates a second or more pairs of student-mentor interviews, the information on the interview card is overwritten. 

**Description**:
Now we render all interview cards, and student can see the results.

image

**Self-Check**:

- [x] Changes tested locally

### DIFF
--- a/client/src/pages/course/student/interviews.tsx
+++ b/client/src/pages/course/student/interviews.tsx
@@ -103,29 +103,27 @@ function StudentInterviewPage() {
               const items = getStudentInterviewItems(name);
 
               if (items.length > 0) {
-                return items.map(item => {
-                  return (
-                    <InterviewCard
-                      key={id + item.id}
-                      interview={interview}
-                      item={item}
-                      isRegistered={true}
-                      onRegister={handleRegister}
-                    />
-                  );
-                });
-              } else {
-                const registered = hasInterview(id);
-                return (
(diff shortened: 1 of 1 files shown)


emp74ark (approved):
It works.
We have two arrays with interviews in one omponent.
It is not obiviouse that  `data` array represents student's interviews and `interviews` array represents possible interviews in the course.

It would be nice to give these variables some explanatory names to understand easily the code in future.

client/src/pages/course/student/interviews.tsx
```diff
@@ -100,17 +100,32 @@ function StudentInterviewPage() {
...
+                    <InterviewCard
+                      key={id + item.id}
+                      interview={interview}
+                      item={item}
+                      isRegistered={true}
+                      onRegister={handleRegister}
+                    />
+                  );
+                });
+              } else {
```
AlreadyBored: `else` branch is not necessary here as you have `return` inside of `if` branch

aliakseitokarev:
> It works. We have two arrays with interviews in one omponent. It is not obiviouse that `data` array represents student's interviews and `interviews` array represents possible interviews in the course.
> 
> It would be nice to give these variables some explanatory names to understand easily the code in future.

@emp74ark done

PR #2890, opened 2025-12-09
fix: adjust pre-screening interview layout for better text wrapping

pestler:
**bug**
image

**solution**
image
**Description**:
adjust pre-screening interview layout for better text wrapping

**Self-Check**:

- [x] Changes tested locally

### DIFF
--- a/client/src/modules/Interview/Student/components/InterviewCard.tsx
+++ b/client/src/modules/Interview/Student/components/InterviewCard.tsx
@@ -30,6 +30,8 @@ export const InterviewCard = ({
   const { id, descriptionUrl, name, startDate, endDate, studentRegistrationStartDate: registrationStart } = interview;
   const interviewPassed = item?.status === InterviewStatus.Completed;
   const interviewResult = getInterviewCardResult(item?.result as Decision);
+  const hasInterviewPair = !!item;
+
   const registrationNotStarted = isRegistrationNotStarted(registrationStart);
   const { cardMessage, backgroundImage } = getInterviewCardDetails({
     interviewResult,
@@ -37,6 +39,7 @@ export const InterviewCard = ({
     isRegistered,
     registrationNotStarted,
     registrationStart,
+    hasInterviewPair,
   });
 
   return (
--- a/client/src/modules/Interview/Student/data/getInterviewCardDetails.tsx
+++ b/client/src/modules/Interview/Student/data/getInterviewCardDetails.tsx
@@ -7,6 +7,7 @@ interface StudentInterviewDetails {
   interviewPassed: boolean;
   registrationStart: string;
   interviewResult: InterviewResult;
+  hasInterviewPair: boolean;
 }
 
 export const getInterviewCardDetails = ({
@@ -15,6 +16,7 @@ export const getInterviewCardDetails = ({
   isRegistered,
   registrationNotStarted,
   registrationStart,
+  hasInterviewPair,
 }: StudentInterviewDetails) => {
   if (interviewPassed) {
     switch (interviewResult) {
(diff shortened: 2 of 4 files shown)


client/src/components/Profile/ui/PrescreeningFeedback.tsx
```diff
@@ -16,10 +16,10 @@ export function PrescreeningFeedback({ feedback }: { feedback: StageInterviewDet
     return (
       <Space direction="vertical">
         {intro.values?.comment && (
-          <Space>
+          <Row>
```
emp74ark: If we just need stack we can stilee use `<Space direction="vertical">`.

`Row` IMHO better looks with `gutter` (`Space` adds it automatically) .

pestler: Done

client/src/components/Profile/ui/PrescreeningFeedback.tsx
```diff
@@ -3,65 +3,109 @@ import { StageInterviewDetailedFeedback } from '@common/models/profile';
...
-export function PrescreeningFeedback({ feedback }: { feedback: StageInterviewDetailedFeedback['feedback'] }) {
-  const { steps } = feedback as { steps: Record<FeedbackStepId, InterviewFeedbackStepData> };
+const STYLES = {
+  feedbackItemWidth: '80%',
+  skillCommentWidth: '60%',
+};
 
-  const { theory, practice, english, decision, intro } = steps;
-  const isRejected = intro.values?.interviewResult === 'missed';
+const FEEDBACK_CONFIG = [
```
emp74ark, line 15: All items in this array have witdh `STYLES.feedbackItemWidth`. The `FeedbackItem` has the same default value for the `width` props. Can we safely skip it here?

client/src/components/Profile/ui/PrescreeningFeedback.tsx
```diff
@@ -3,65 +3,109 @@ import { StageInterviewDetailedFeedback } from '@common/models/profile';
...
+
+export function PrescreeningFeedback({ feedback }: { feedback: StageInterviewDetailedFeedback['feedback'] }) {
+  const { steps } = feedback as { steps: Record<FeedbackStepId, InterviewFeedbackStepData> };
+  const { theory, practice } = steps;
+
+  const isRejected = steps.intro.values?.interviewResult === 'missed';
+
+  const displayItems = React.useMemo(
+    () =>
+      FEEDBACK_CONFIG.filter(item => item.isRejectedItem === isRejected).map(item => ({
```
emp74ark: Why do we need to add `isRejected` items to the `FEEDBACK_CONFIG` and filter by this parameter here? Do we need rejected items anywhere else?

client/src/components/Profile/ui/PrescreeningFeedback.tsx
```diff
@@ -3,65 +3,109 @@ import { StageInterviewDetailedFeedback } from '@common/models/profile';
...
+        id: item.id,
+        label: item.label,
+        value: item.getValue(steps),
+        width: item.width,
+      })),
```
emp74ark: The same:
```tsx
FEEDBACK_CONFIG.filter(item => item.isRejectedItem === isRejected).map(({ id, label, getValue, width }) => ({ id, label, value: getValue(steps), width })
)
```

client/src/components/Profile/ui/PrescreeningFeedback.tsx
```diff
@@ -3,65 +3,109 @@ import { StageInterviewDetailedFeedback } from '@common/models/profile';
...
-          </Space>
-        )}
-        <SkillSection skills={theory.values} title="Theory" tooltips={SKILLS_LEVELS} />
-        <SkillSection skills={practice.values} title="Practice" tooltips={CODING_LEVELS} />
-      </Space>
-    </>
+    <Space direction="vertical" size={20}>
+      {displayItems.map(item => (
+        <FeedbackItem key={item.id} label={item.label} value={item.value} width={item.width} />
+      ))}
```
emp74ark, line 95: Doesn't really matter, but the same:
```tsx
{displayItems.map(({ id, label, value, width }) => (
  <FeedbackItem key={id} label={label} value={value} width={width} />
))}
```

client/src/components/Profile/ui/PrescreeningFeedback.tsx
```diff
@@ -3,65 +3,103 @@ import { StageInterviewDetailedFeedback } from '@common/models/profile';
...
+
+export function PrescreeningFeedback({ feedback }: { feedback: StageInterviewDetailedFeedback['feedback'] }) {
+  const { steps } = feedback as { steps: Record<FeedbackStepId, InterviewFeedbackStepData> };
+  const { theory, practice } = steps;
+
+  const isRejected = steps.intro.values?.interviewResult === 'missed';
+
+  const displayItems = React.useMemo(
+    () =>
+      FEEDBACK_CONFIG.filter(item => item.isRejectedItem === isRejected).map(({ id, label, getValue }) => ({
```
AlreadyBored: to double-check: is it correct that we filter by isRejected (which can be both `true` or `false`) instead of filtering by specific value for `item.isRejectedItem`?

client/src/components/Profile/ui/PrescreeningFeedback.tsx
```diff
@@ -3,65 +3,103 @@ import { StageInterviewDetailedFeedback } from '@common/models/profile';
 import { CODING_LEVELS, FeedbackStepId, SKILLS_LEVELS } from 'data/interviews/technical-screening';
 import { InterviewFeedbackStepData, InterviewFeedbackValues, InterviewQuestion } from '@common/models';
 import { Rating } from '@client/components/Rating';
+import React from 'react';
```
AlreadyBored: Let's import only what we need  
```suggestion
import { useMemo } from 'react';
```

PR #2910, opened 2026-01-24
feat: client devtools

emp74ark:
**Description**:
This PR introduces a new feature: **Client DevTools** - an extensible client-side module designed to simplify routine operations and support development.

Currently, it includes the ability to log in as a user from a predefined list without manually editing the `.env` file.

The Client DevTools components depend on the NestJS **DevTools** module.

All new modules and components are available **only in non-production (development) environments**.

**Screencast**
[Screencast_20260204_232908.webm](https://github.com/user-attachments/assets/0414f42a-6a95-4947-9589-5b6a9e4baa90)

**Self-Check**:
- [x] Changes tested locally

### DIFF
--- a/client/src/components/DevTools/DevToolsCurrentUser.tsx
+++ b/client/src/components/DevTools/DevToolsCurrentUser.tsx
@@ -24,16 +24,11 @@ export default function DevToolsCurrentUser() {
   }
 
   return (
-    <Descriptions
-      title="Active user session"
-      layout="vertical"
-      bordered={true}
-    >
+    <Descriptions title="Active user session" layout="vertical" bordered={true}>
       <Item label="User ID">{data?.id}</Item>
       <Item label="GitHub ID">{data?.githubId}</Item>
       <Item label="Admin">{data?.isAdmin ? 'true' : 'false'}</Item>
-      <Item label="Course ids">{Object.keys(data?.courses ?? {}).
-        toString()}</Item>
+      <Item label="Course ids">{Object.keys(data?.courses ?? {}).toString()}</Item>
     </Descriptions>
   );
 }
--- a/client/src/pages/_app.tsx
+++ b/client/src/pages/_app.tsx
@@ -6,11 +6,7 @@ import 'antd/dist/reset.css';
 import { initializeFeatures } from 'services/features';
 import { Analytics } from '../components/Analytics';
 import '../styles/main.css';
-import {
-  DevToolsProvider,
-  MessageProvider,
-  ThemeProvider,
-} from '@client/providers';
+import { DevToolsProvider, MessageProvider, ThemeProvider } from '@client/providers';
 
 class RsSchoolApp extends App {
   render() {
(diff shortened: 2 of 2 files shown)


nestjs/src/ormconfig.ts
```diff
@@ -17,7 +17,7 @@ const config: DataSourceOptions = {
   database: process.env.RSSHCOOL_PG_DATABASE,
   entities: models,
   migrations,
-  synchronize: false,
+  synchronize: process.env.NODE_ENV !== 'production',
```
AlreadyBored: Why? We control DB schema changes via migrations, this change may lead to the situation when you'll forget to write migration and your code will fail in prod + migrations are applied on `server` start, what's the point of making `synchronize: true`? I see only risks, no benefits

emp74ark: Removed

nestjs/src/devtools/devtools.module.ts
```diff
@@ -0,0 +1,21 @@
...
+import { DevtoolsService } from './devtools.service';
+import { User } from '@entities/user';
+import { ConfigModule } from '../config';
+
+@Module({})
+export class DevtoolsModule {
+  static forRoot(): DynamicModule {
+    const isDev = process.env.NODE_ENV !== 'production';
+    const staging = process.env.RS_ENV === 'staging';
+    const devToolsAllowed = !staging && isDev;
```
AlreadyBored: So we don't allow devtools in feature deploys, right?

nestjs/src/devtools/devtools.service.ts
```diff
@@ -0,0 +1,48 @@
...
+        mentors: {
+          courseId: true,
+          course: {
+            alias: true,
+          },
+        },
+      },
+      relations: ['students.course', 'mentors.course'],
+    });
+    console.log(users);
```
AlreadyBored: Fogot to remove or we need this for some purpose?

emp74ark: Fixed

nestjs/src/devtools/devtools.module.ts
emp74ark: Updated. Please check again.

client/src/components/DevTools/DevToolsCurrentUser.tsx
```diff
@@ -0,0 +1,35 @@
...
+  return (
+    <Flex vertical>
+      <ul>
+        <li>User ID: {data.id}</li>
+        <li>GitHub ID: {data.githubId}</li>
+        <li>Admin: {data.isAdmin ? 'true' : 'false'}</li>
+        <li>Courses: {Object.keys(data.courses ?? {}).toString()}</li>
+      </ul>
+    </Flex>
+  );
```
AlreadyBored, line 33: Please consider use Antd Descriptions for this https://ant.design/components/descriptions

client/src/components/DevTools/DevToolsUsers.tsx
```diff
@@ -0,0 +1,90 @@
...
+  async function loginWithUser(user: DevtoolsUserDto) {
+    try {
+      await devToolsApi.getDevUserLogin(user.githubId);
+      router.push('/api/v2/auth/github/login');
+    } catch {
+      console.error('Failed to login user');
+    }
+  }
+
+  const tableColumns = [
```
AlreadyBored: Please consider add `useMemo` here

client/src/providers/Providers.tsx
AlreadyBored, line 1: This adds unnecessary complexity, let's simply add new provider at client/src/pages/_app.tsx and get rid of this file

emp74ark: Updated

PR #2948, opened 2026-03-12
fix: add vitest config to CI workspace

AlreadyBored:

**Issue**:
N/A

**Description**:
Fixes issue with running test job in CI for server after migration to `vitest`  
https://github.com/rolling-scopes/rsschool-app/actions/runs/22978655332/job/66713251101  
image  
Root cause: 
The issue was in `deploy.yaml` — the workflow uses `turbo prune --scope=server --docker` to create an isolated workspace in `app/`. The root-level `vitest.shared.ts` file is not included in the pruned output, while `server/vitest.config.ts` imports it via `import shared from '../vitest.shared'`.

**Fix:** added `cp vitest.shared.ts app/` to the copy steps for both jobs — `deploy_build_server` and `deploy_build_nestjs` — so that `vitest.shared.ts` is available in the isolated workspace.

**Self-Check**:

- [ ] Changes tested locally

### DIFF
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -10,11 +10,47 @@ env:
   TURBO_VERSION: '2.8.12'
 
 jobs:
-  deploy_build_client:
+  checks:
+    name: Test, Lint
+    concurrency:
+      group: deploy-checks
+      cancel-in-progress: true
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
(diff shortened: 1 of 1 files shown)


apalchys:
hmm... we do it wrong. We should run test, lint, format and onlt after that generate app specific resources for Docker

apalchys:
If you don't mind, I will take this branch and fix it properly later today

PR #2953, opened 2026-03-15
refactor: replace custom `useLoading` hook with `ahooks` `useRequest`

apalchys:
### Motivation
- Remove custom loading wrapper and standardize async request handling with `ahooks` `useRequest` for clearer semantics and built-in error/side-effect handling.
- Consolidate loading state management and reduce duplicated boilerplate around try/catch/finally patterns.

### Description
- Deleted the custom hook file `client/src/components/useLoading.tsx` and replaced its usage with `useRequest` from `ahooks` across multiple modules. 
- Converted `const [loading, withLoading] = useLoading()` call sites into `const { loading, runAsync: fn } = useRequest(..., { manual: true })` or aggregated multiple `useRequest` loadings into a combined `loading` flag where appropriate.
- Added `onError` handlers to several `useRequest` calls to surface user-facing messages via `message.error` and preserved previous error-handling behavior in most places.
- Updated many pages and components (for example under `Mentor`, `Teams`, `Students`, `Interviews`, `Notifications`, `Opportunities`, `MentorRegistry`, `AutoTest`, `StudentDashboard`, and others) to call the new `runAsync` functions and to use `useRequest` for manual triggers and refresh dependencies.


client/src/modules/Opportunities/pages/EditPage/index.tsx
```diff
@@ -29,29 +28,37 @@ export function EditPage() {
...
+    },
+    {
+      manual: true,
+      onError: () => {
+        message.error('An unexpected error occurred. Please try later.');
+      },
+    },
+  );
 
-  const handleDeleteConsent = withLoading(async () => {
```
AlreadyBored, line 50: These two funcs lost their `withLoading` wrapping entirely. No loading state, no error handling, is it okay?

client/src/pages/admin/mentor-registry.tsx
```diff
@@ -83,53 +81,65 @@ function Page() {
...
-    try {
-      await mentorRegistryService.sendCommentMentorRegistry(modalData?.record?.githubId, comment);
-      await loadData();
-    } catch {
-      message.error('An error occurred. Please try again later.');
-    } finally {
+  const { loading: loadingCancelMentor, runAsync: cancelMentor } = useRequest(
+    async (githubId: string) => {
+      setModalData(null);
+      await mentorRegistryService.cancelMentorRegistry(githubId);
```
AlreadyBored, line 125: No `onError` here — if the request fails, the user gets no feedback.

AlreadyBored (commented):
`useLoading` swallowed errors — callers never saw exceptions, just got `undefined` back. `runAsync` re-throws after `onError`, so any call site without `try/catch` will now produce an unhandled promise rejection. Is that what we want?

AlreadyBored (approved):

PR #2996, opened 2026-05-13
fix(mentor): translate unique_violation to 409 on concurrent mentor registration

calebcgates:

**Issue**:
_N/A — surfaced during a TypeORM error-handling audit of `server/src/routes/`. No upstream issue. Happy to open one first if preferred per repo convention._

**Description**:

`server/src/routes/course/mentor.ts:83` calls `mentorRepository.insert({ courseId, userId, ...data })` without local error handling. The `(courseId, userId)` pair is enforced by a UNIQUE index on the `mentor` table, so two concurrent `POST /:courseId/mentor/:githubId` requests for the same user — racing past the `findOne(...)` check on line 66 — both pass the "does this mentor already exist?" guard and both attempt `INSERT`. The loser hits the unique-constraint violation, which TypeORM raises as `QueryFailedError` with `driverError.code === '23505'`.

Today that error propagates to the global `errorHandlerMiddleware` (`server/src/routes/logging.ts:13`), which logs it and returns HTTP 500 with a generic Postgres error string in the body (`duplicate key value violates unique constraint "..."`). The client sees a 500 for what is genuinely a client-recoverable conflict (the mentor IS registered — just by the other concurrent request).

This PR adds a local try/catch around the `insert(...)` call, narrows on `QueryFailedError` + the new `isUniqueViolation(err)` helper, and translates the race to HTTP 409 Conflict with `{ message: 'Mentor is already registered for this course (concurrent request).' }`. All other errors (including `QueryFailedError` with non-23505 codes) re-throw to the existing global handler — no swallowing.

The helper `isUniqueViolation(error: QueryFailedError): boolean` is added to `server/src/routes/utils.ts` so other route handlers that wrap `repo.insert()` / `repo.save()` calls can reuse it. The check is minimal (`driverError?.code === '23505'`) and matches Postgres' `unique_violation` error class.

**Self-Check**:

- [ ] Database migration added (if required) — N/A (no schema change)
- [x] Changes tested locally

### Local verification

```sh
cd server
npx vitest run src/routes/course/__test__/mentor.test.ts   # 3/3 passing
npm run compile                                            # 0 type errors
npx eslint src/routes/course/mentor.ts src/routes/utils.ts src/routes/course/__test__/mentor.test.ts   # 0 issues
npx oxfmt --check server/src/routes/course/mentor.ts server/src/routes/utils.ts server/src/routes/course/__test__/mentor.test.ts   # passes
npx vitest run                                              # full server suite: 6 files, 36/36 passing
```

The new test covers the `isUniqueViolation` helper across three branches: 23505 → true, a different SQL state (23502 not_null_violation) → false, missing driverError code → false. Happy to add a route-level integration test (mock `mentorRepository.insert` to throw, assert response status + body) if you'd prefer end-to-end coverage; it'd be one extra test file.

### DIFF
--- a/server/src/routes/course/__test__/mentor.test.ts
+++ b/server/src/routes/course/__test__/mentor.test.ts
@@ -1,0 +1,20 @@
+import { describe, it, expect } from 'vitest';
+import { QueryFailedError } from 'typeorm';
+import { isUniqueViolation } from '../../utils';
+
+describe('isUniqueViolation', () => {
+  it('returns true for postgres unique_violation (23505)', () => {
+    const err = new QueryFailedError('insert', [], { code: '23505' } as unknown as Error);
+    expect(isUniqueViolation(err)).toBe(true);
+  });
+
+  it('returns false for a different postgres error (23502 not_null_violation)', () => {
+    const err = new QueryFailedError('insert', [], { code: '23502' } as unknown as Error);
+    expect(isUniqueViolation(err)).toBe(false);
+  });
+
+  it('returns false when driverError has no code', () => {
+    const err = new QueryFailedError('insert', [], {} as Error);
+    expect(isUniqueViolation(err)).toBe(false);
+  });
+});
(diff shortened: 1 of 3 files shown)


AlreadyBored (approved):

AlreadyBored:
@calebcgates thanks!

Issue #2994, opened 2026-05-05
Issue with Checkbox


TeymurDev:
**Describe the bug**
Checkbox is not fully visible.

**To Reproduce**

**Expected behavior**
Checkbox is fully visible

**Screenshots**

<img width="1076" height="472" alt="Image" src="https://github.com/user-attachments/assets/f49da0d2-afb3-415b-b2d6-b1cc878f7fe9" />

**Device (please complete the following information):**

**Additional context**

PR #2997, opened 2026-05-13
fix: fix ant inputs shrink

DenHelloWorld:
**Description**:
This PR fixes the layout issue where checkboxes and radio buttons were being compressed in flex containers, as reported in #2994.

### The Problem:
Inside flexbox layouts, checkboxes and radio buttons could shrink when the label text was long or the container width was limited. This caused visual distortion where the selection components would lose their proportions and appear "squashed".

### The Solution:
- Added `flex-shrink: 0;` to `.ant-checkbox` and `.ant-radio` styles.
- This ensures that these elements always maintain their fixed dimensions regardless of the flex container's space distribution.

**Screenshots**:
| Before | After |
| :--- | :--- |
| <br> | <br> |

**Self-Check**:
- [x] Changes tested locally


AlreadyBored:
@DenHelloWorld  thank you!

AlreadyBored (approved):

Issue #3001, opened 2026-06-08
Custom certificate template


AlreadyBored:
**Describe the feature**

Allow a Course Manager to choose a certificate **template** when issuing certificates, instead of always using the single hard-coded design. The selected template defines the visual look of the rendered PDF (background, layout, fonts); student / course / issuer data stay the same.

Scope of this first iteration (hard-coded templates — no admin UI, no DB table):

- A small catalog of predefined templates:
  - `default` — current RS School certificate (black / yellow);
  - `bootcamp_13_weeks` — "JS/Frontend Summer Bootcamp" design (capybara, navy / gold).
- Template can be picked in **both** issuance flows:
  - batch issuance — in the Certificate Criteria modal;
  - single issuance — when issuing a certificate to one student.
- Each option shows a **preview thumbnail**, plus a fullscreen icon to open an almost full-screen preview so the certificate details are readable before issuing.
- Access stays as-is (Course Manager) — no permission changes.

Out of scope (possible follow-ups): per-course default template, storing the chosen template on the `Certificate` entity, user-uploadable templates, a template-management admin page.

**Additional context**

- Rendering lives in the private AWS repo: a Lambda renders an EJS/HTML template to PDF via Puppeteer and uploads it to S3. Templates are stored as subfolders (`certificate-templates/<id>/`) and selected by a `templateId` passed through the request → SQS → renderer.
- The public app (`rsschool-app`) sends `templateId` from the UI down to the AWS REST API; unknown / missing values fall back to `default`.
- No database migration is required for this iteration.

PR #3002, opened 2026-06-08
feat: allow custom certificate template

AlreadyBored:

**Issue**:
#3001 

**Description**:

Adds the ability to choose a certificate **template** when issuing certificates (both single and batch), available to Course Managers. The selected `templateId` is propagated end-to-end to the certificate renderer; student / course / issuer data are unchanged.

This iteration ships two hard-coded templates — `default` and `bootcamp_13_weeks` — each with a preview thumbnail and an almost full-screen preview (fullscreen icon → zoomable overlay) so the design can be inspected before issuing. There is a companion PR in AWS repo that adds the matching template folders and renderer support — **both must be deployed together**.

Changes:

- **Catalog (single source of truth)** — `nestjs/src/certificates/templates/catalog.ts`: `CERTIFICATE_TEMPLATES` (id / label / previewUrl) + `resolveCertificateTemplateId()` with a whitelist fallback to `default`.
- **API** — new `GET /certificate/templates` (`nestjs/.../certificates.controller.ts`) returning the catalog for the UI.
- **Server** — `server/src/routes/course/certificates.ts`: both `postCertificates` (batch) and `postStudentCertificate` (single) read, validate, and forward `templateId` in the payload sent to the AWS REST API (invalid / missing → `default`).
- **Client**:
  - `client/src/services/course.ts` — `createCertificate(githubId, templateId)` and `postCertificateStudents(criteria, templateId)`.
  - New `CertificateTemplatePicker` (`.tsx` + `.module.css`) — thumbnails with a radio selection and a fullscreen preview (Antd `Image`, opened ~88vh via `imageRender`).
  - New `IssueCertificateModal` for single issuance.
  - `CertificateCriteriaModal.tsx` — added a template field for batch issuance.
  - `students.tsx` / `DashboardDetails.tsx` — wired single issuance through the new modal and pass `templateId` in batch.
  - `components/index.ts` — exports for the new components.
  - Preview thumbnails (2048px): `client/public/static/certificate-templates/{default,bootcamp_13_weeks}.jpg`.

Permissions are unchanged (Course Manager). No `Certificate` entity / DB schema changes.

**Self-Check**:

- [ ] Database migration added (if required) — _not required; no schema changes_
- [x] Changes tested locally


client/src/modules/CourseManagement/components/CertificateTemplatePicker/CertificateTemplatePicker.tsx
```diff
@@ -0,0 +1,135 @@
...
+      .finally(() => {
+        if (!cancelled) setLoading(false);
+      });
+    return () => {
+      cancelled = true;
+    };
+  }, []);
+
+  useEffect(() => {
+    if (value == null && templates.length > 0) {
```
emp74ark: Due to value has type `string | undefined` probably better use `value === undefined`

client/src/modules/CourseManagement/components/CertificateTemplatePicker/CertificateTemplatePicker.tsx
```diff
@@ -0,0 +1,135 @@
...
+}
+
+export function CertificateTemplatePicker({ value, onChange }: Props) {
+  const [templates, setTemplates] = useState<Template[]>(cachedTemplates ?? []);
+  const [loading, setLoading] = useState(cachedTemplates == null);
+  const [previewSrc, setPreviewSrc] = useState<string | undefined>(undefined);
+
+  useEffect(() => {
+    let cancelled = false;
+    if (cachedTemplates) {
```
emp74ark, line 49: it is always `true` due to default value `[]` in `useState`

client/src/modules/CourseManagement/components/CertificateTemplatePicker/CertificateTemplatePicker.tsx
```diff
@@ -0,0 +1,135 @@
...
+      cancelled = true;
+    };
+  }, []);
+
+  useEffect(() => {
+    if (value == null && templates.length > 0) {
+      const fallback = templates.find(t => t.id === DEFAULT_TEMPLATE_ID)?.id ?? templates[0]?.id;
+      if (fallback) onChange?.(fallback);
+    }
+  }, [templates, value, onChange]);
```
emp74ark: `onChange` in deps looks risky because the parent component do not memoised `setTemplateId`.

client/src/modules/CourseManagement/components/CertificateTemplatePicker/CertificateTemplatePicker.tsx
```diff
@@ -0,0 +1,135 @@
...
+          src: previewSrc,
+          onOpenChange: open => {
+            if (!open) setPreviewSrc(undefined);
+          },
+          // Open the preview almost full-screen so the certificate details are readable.
+          imageRender: originalNode => {
+            if (!isValidElement(originalNode)) return originalNode;
+            const node = originalNode as ReactElement<{ className?: string }>;
+            return cloneElement(node, {
+              className: [node.props.className, styles.previewImg].filter(Boolean).join(' '),
```
emp74ark: What about clsx: `clsx(node.props.className, styles.previewImg)`?

client/src/modules/CourseManagement/components/CertificateTemplatePicker/CertificateTemplatePicker.tsx
```diff
@@ -0,0 +1,135 @@
...
+              </div>
+            </Radio>
+          );
+        })}
+      </Radio.Group>
+
+      <Image
+        classNames={{ root: styles.hidden }}
+        preview={{
+          open: previewSrc != null,
```
emp74ark: `open: !previewSrc`? previewSrc has type `string | undefined`

client/src/modules/CourseManagement/components/CertificateTemplatePicker/CertificateTemplatePicker.tsx
```diff
@@ -0,0 +1,135 @@
...
+          src: previewSrc,
+          onOpenChange: open => {
+            if (!open) setPreviewSrc(undefined);
+          },
+          // Open the preview almost full-screen so the certificate details are readable.
+          imageRender: originalNode => {
+            if (!isValidElement(originalNode)) return originalNode;
```
emp74ark: is it possible to have an invalid originalNode?[ AntDesign docs](https://ant.design/components/image) says, that it always has type `originalNode: React.ReactElement`

client/src/modules/CourseManagement/components/CertificateTemplatePicker/CertificateTemplatePicker.tsx
AlreadyBored: Agreed — since the prop is typed `string | undefined`, strict `=== undefined` is more precise. Fixed in 02c8b241.

client/src/modules/CourseManagement/components/CertificateTemplatePicker/CertificateTemplatePicker.tsx
AlreadyBored: It's not always `true` — `cachedTemplates` here is the module-level cache (`Template[] | null`), not the `templates` state, so it stays `null` until the first successful fetch. That said, the branch body was indeed redundant: the `useState` initializers above already seed state from the cache, so I simplified it to an early return in 02c8b241.

client/src/modules/CourseManagement/components/CertificateTemplatePicker/CertificateTemplatePicker.tsx
AlreadyBored: `setTemplateId` is actually stable — React guarantees identity of `useState` setters. The riskier consumer is the antd Form-injected `onChange` (in `CertificateCriteriaModal` the picker sits inside `Form.Item`), which isn't guaranteed to be memoized. Reworked in 02c8b241: the latest callback is kept in a ref, so the effect no longer depends on callback identity at all.

client/src/modules/CourseManagement/components/CertificateTemplatePicker/CertificateTemplatePicker.tsx
AlreadyBored: Good call — `clsx` is already a client dependency and used elsewhere in the project. Switched (here and for the selected-card class) in 02c8b241.

client/src/modules/CourseManagement/components/CertificateTemplatePicker/CertificateTemplatePicker.tsx
AlreadyBored: `!previewSrc` would invert the logic — the preview would be "open" exactly when there is no src. I assume you meant boolean coercion, so switched `previewSrc != null` → `!!previewSrc` in 02c8b241.

client/src/modules/CourseManagement/components/CertificateTemplatePicker/CertificateTemplatePicker.tsx
AlreadyBored: True — antd types `originalNode` as `React.ReactElement`, so the runtime guard was dead code. Removed it in 02c8b241 (kept only the narrow cast needed to read `className`).

emp74ark (approved):

Issue #1123, opened 2022-01-16
Tech Debt: Migrate NestJS to Fastify


apalchys:
Fastify is much faster than Express and we want to get that performance improvement

- [ ] Switch NestJS to Fastify
- [ ] Replace Passport with something different
- [ ] Update Swagger

PR #3098, opened 2026-08-10
chore(nestjs): upgrade to NestJS 11 (Express 5, cache-manager 6)

AlreadyBored:
## Summary

Phase 3 of #1123: Nest core 10 → 11, so the upcoming adapter swap lands on **supported fastify 5** (platform-fastify@10 would pin us to EOL fastify 4). Half the `@nestjs/*` ecosystem here already shipped Nest-11-targeting majors (passport 11, typeorm 11, config 4, schedule 5, event-emitter 3, axios 4) — this PR closes that split.

| package | before → after |
|---|---|
| @nestjs/common, core, platform-express, testing, cli, schematics | 10.4.x → **11.1.29** |
| @nestjs/swagger | 8.1.1 → **11.4.6** |
| @nestjs/cache-manager / cache-manager | 2.3.0 / 5.7.6 → **3.1.3 / 6.4.3** |
| express / @types/express | 4.21.2 / 4 → **5.2.1 / 5** |

## Notable decisions

- **Root `overrides` for `@nestjs/{common,core,platform-express}`** — `@openapitools/openapi-generator-cli` (devDep) depends on `@nestjs/axios`, and npm insisted on materializing a second `@nestjs/common@10` subtree for it. Two `common` copies in one process break exception-class identity across packages — reproduced as the smoke suite's 401s turning into 500s (`UnauthorizedException` from one copy isn't an `HttpException` to the other's filter). The override collapses the whole tree to a single 11.1.29 instance; `npm ls @nestjs/common` now shows exactly one copy.
- **`package-lock.json` regenerated from scratch** — incremental installs kept resurrecting orphaned v10 nodes from the old lock. Ranged deps float within their existing ranges; both workspaces' suites are the verification.
- **cache-manager 2→3 / 5→6**: TTL stays in milliseconds in both lines, `get/set/del` signatures unchanged — the load-bearing 10-minute auth-session cache and the 11 `CacheInterceptor` endpoints behave identically (covered by unit + smoke tests).
- **Express 5 fallout**: none left — the `'/*'` middleware wildcards were removed in #3088 ahead of time; a route-pattern audit found no optional/regex path params (path-to-regexp v8 safe); `express.json` import in setup.ts works unchanged.

## Verification

- nestjs: full `test:ci` green — **2073 tests / 185 files**, coverage ~99.7% (gate 90). The 4 smoke tests that failed mid-upgrade were the dual-common artifact above; green on the deduped tree.
- `tsc` on `src/`: 0 errors.
- **Live boot on Express 5**: `/swagger` and `/swagger-json` 200, public endpoint 200, 401/404 body shapes unchanged, `Cache-Control: no-cache` on GET, pino request logging intact.
- `spec.json`: swagger 11 adds explicit `"type": "object"` to nullable `allOf` schemas (more correct OpenAPI); **regenerated client SDK is byte-identical**.
- client: local run of the suite kept getting killed by session restarts — relying on CI here; the lockfile regen is the only client-affecting change (no client source touched).

Refs #1123. Next up after this merges: the platform-fastify@11 swap PR (small, revertible), with the smoke harness flipped to dual-adapter mode.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


AlreadyBored:
CI failure root-caused: the from-scratch lockfile regen floated the client's ranged deps and the client suite started hanging (the 40m job cancel). Pushed a fix: the lockfile is now master's with only the `@nestjs/*` + `nestjs-pino` entries dropped and re-resolved under the overrides — single `@nestjs/common@11.1.29` in the tree, and a per-package version diff against master confirms **zero changes to client-runtime packages** (only the express-5 family, swagger-ui-dist and nestjs tooling moved). Verified locally after the fix: nestjs `test:ci` green (2073 tests, coverage gate), client `test:ci` green (full run, coverage gate).

PR #3115, opened 2026-09-11
feat: add branded domain identifiers

apalchys:
### Motivation

- Introduce nominal (branded) ID types to improve type-safety and avoid unchecked `as` casts for high-use identifiers. 
- Perform runtime validation for constructors so numeric IDs are safe positive integers and UUID IDs match the expected version. 
- Apply branded IDs incrementally at controller → service boundaries to avoid invasive repository changes while keeping TypeORM columns compatible with DB primitives.

### Description

- Add `nestjs/src/core/types/brand.ts` providing a generic `Brand<Value, Name>` type, an explicit `brandValue` constructor and `serializeBrand` helper. 
- Add `nestjs/src/core/types/identifiers.ts` with distinct branded types and runtime constructors for numeric IDs (`toUserId`, `toCourseId`, etc.) and UUID v4 IDs (`toResumeId`, `toCourseLeaveSurveyResponseId`) plus serialization helpers. 
- Add unit tests next to the types in `nestjs/src/core/types/brand.test.ts` and `nestjs/src/core/types/identifiers.test.ts` covering valid construction and invalid edge-cases (zero, negative, fractional, unsafe integers, malformed/incorrect UUIDs). 
- Convert controller primitives into validated branded IDs at entry points for `courses`, `registry`, `certificates`, and `auth` and update service method signatures to accept the matching branded types; keep TypeORM column primitives but type relation IDs with the brands where safe (example: `Certificate.studentId: StudentId`).


nestjs/src/core/types/brand.ts
```diff
@@ -0,0 +1,9 @@
+declare const brand: unique symbol;
+
+export type Brand<Value, Name extends string> = Value & {
+  readonly [brand]: Name;
+};
+
+export const brandValue = <Value, Name extends string>(value: Value): Brand<Value, Name> => value as Brand<Value, Name>;
+
+export const serializeBrand = <Value>(value: Value): Value => value;
```
AlreadyBored, line 9: why do we need this function, it doesn't change neither type of a value or value itself

